PR #2414, opened 2014-12-18
Added Plugin::onError() to handle plugin exceptions

PEMapModder:
I called it Error instead of Exception because some class names might therefore end in "Exception" and lead to confusion that they are subclasses of \Exception.

This change calls the Plugin::onError() method when an exception is thrown from a plugin command execution, event execution or a plugin task execution. This enables plugins to have features like auto-reporting exceptions to their authors or create other custom handlers.


shoghicp:
This is being closed due to not being implemented.

PR #2485, opened 2015-01-02
Makes a Folder to keep all the CrashDump.logs

MegaSamNinja:
Not sure if correct
probably not correct


shoghicp:
No, it shouldn't crash that much. This last issue was unexpected, should be ok now. Also, this will only affect servers on a restart loop, that is an user issue.

PEMapModder:
also @MegaSamNinja the syntax is wrong, the last `.`

PR #2540, opened 2015-01-14
Add Flowers

MrPowerGamerBR:
Now player's houses can be more fancier than ever before!

Allows Placing of Blue Orchids, Alliums, Azure Bluet, all kinds of Tulips and Oxeye Daisys (Except double block flowers)

As for now, it will not generate in worlds.

(Sorry if there is some error in the file, first time editing PocketMine-MP source)


shoghicp:
This is not the proper way to implement them. Please read the Contribution Guidelines.

PR #2573, opened 2015-01-27
Added a Time in sec. for ban.

seria2:


PEMapModder:
1. Please follow the format as specified in the contributing guidelines.
2. You disallowed reasons with multiple words.
3. This should be done via a plugin.

PR #2574, opened 2015-01-27
Implemented hardcore banning

LDX-MCPE:
I'm not sure if I should kick the player or not.


sekjun9878:
Hmm.. interesting. If the player is dead then you should kick the player otherwise he could just respawn? Or is the respawn blocked somewhere

PEMapModder:
No, you should kick the player.

https://github.com/PocketMine/PocketMine-MP/blob/master/src/pocketmine/Player.php#L2024-2027

From the above link, we can see that PocketMine does not seem to detect if a player is banned when he dies.

PEMapModder:
https://github.com/LDX-MCPE/PocketMine-MP/pull/1

LDX-MCPE:
I wondered about that. That's why I said:
> I'm not sure if I should kick the player or not.

PR #2596, opened 2015-02-01
Fixed wrong time value

Sulfatezzz:
See http://minecraft.gamepedia.com/Day-night_cycle


shoghicp:
See http://minecraft.gamepedia.com/Day-night_cycle
> /time set night || Sets the time to 14000.

PR #2608, opened 2015-02-04
Fixed for Iron bars (It drops after using hands to break it)

mmlmml1:
image


raginruffalo:
They aren't supposed to drop when you use your hands, only when you use a PickAxe.

mmlmml1:
I mean after this, it will not drop.

raginruffalo:
Oh OK I was under the impression that you were making then drop after it
being disabled previously.

PR #2616, opened 2015-02-06
Added function to align lines with MCPE font

PEMapModder:
A shorthand method for doing this in `Player::setNametag()` was added.

* Backwards-compatible
* Recommended tags: `PR: Contribution`

API Changes:
* Changed `public function pocketmine\Player::setNametag(string $name)` to `public function pocketmine\Player::setNametag(string $name, int $align = pocketmine\utils\TextWrapper::ALIGN_LEFT` (it will _not_ affect calls expecting older versions)
* Added `public static function pocketmine\utils\TextWrapper::align(string[] $lines, int $mode) throws InvalidArgumentException`
* Added `public static function pocketmine\utils\TextWrapper::getLength(string $line, int $utfValue = 8)`

Widths of characters not included in `pocketmine\utils\TextWrapper::$allowedChars` are assumed as 8.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -488,12 +488,12 @@ class Player extends Human implements CommandSender, InventoryHolder, IPlayer{
 	public function getNameTag(){
 		return $this->nameTag;
 	}
-
 	/**
 	 * @param string $name
+	 * @param int    $align default TextWrapper::ALIGN_LEFT
 	 */
-	public function setNameTag($name){
-		$this->nameTag = $name;
+	public function setNameTag($name, $align = TextWrapper::ALIGN_LEFT){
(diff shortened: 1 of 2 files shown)


LDX-MCPE:
:+1:

Falkirks:
"Center" is US and "Centre" is outside US. I have only seen "Center" used in programming.

shoghicp:
This might not be needed soon.

PEMapModder:
But right now?

And what about the getLength() function? It is not possible to get the length from a plugin without using reflections or copying the numbers; and copying them doesn't seem to be a good idea, because this should be done by an API instead.

shoghicp:
Text wrapping has been added [see this tweet](https://twitter.com/_tomcc/status/567311965945868288). Aligning fonts will break on different devices and places, and should not be done automatically.

PR #2618, opened 2015-02-07
Update PigZombie.php

FatalDeath:


PEMapModder:
Why are there so many empty lines before the EOF? To get contribution lines?

And right now this is useless unless you use it somewhere.

sekjun9878:
Please remove the extra newlines before this can be considered for merge to master.

Falkirks:
@FatalDeath also when you remove the new lines it would be optimal if you squash the commits, that way there isn't a commit on master that is just removing line breaks :smile:

shoghicp:
Mobs won't be partially implemented. This was done for Zombies and Villagers, and it was a bad idea.

PR #2620, opened 2015-02-07
Update start.sh

adrianwebster:
Add wait for 10 seconds

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -29,6 +29,8 @@ if [ "$PHP_BINARY" == "" ]; then
 		PHP_BINARY=$(type -p php)
 	else
 		echo "Couldn't find a working PHP binary, please use the installer."
+		echo "Exiting in 10 seconds"
+		sleep 10
 		exit 1
 	fi
 fi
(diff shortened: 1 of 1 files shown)


64FF00:
O_o?

adrianwebster:
It adds wait for ten seconds (sleep) before it exits for user to be able to read error without instantly closing

adrianwebster:
I just realized i forgot to do both...

ns827:
Linux terminal emulator does not close the window automatically when the program exits. I don't know why this patch is useful.

Falkirks:
@ns827 Doesn't the windows shell close out?

ns827:
@Falkirks I don't know because I use only Ubuntu, but I remember mintty prints segmentation fault message on 'normal' server stop.

\+ This pull request is about bash shell scripts, not windows batch.

adrianwebster:
Ok nvm just dont merge it I understand now lol i never had the error but i just assumed it closed but it doesnt xD

64FF00:
@ns827 :+1:

PR #2631, opened 2015-02-12
TallGrass blocks should be only placed on grass blocks

64FF00:
... Maybe not?


Yosshi999:
TallGrass is 2blocks high so i think you should add more functions.

64FF00:
?

iksaku:
:+1:

PR #2634, opened 2015-02-14
Fixed #2615

### LINKED ISSUE #2615: torches cannot be placed on cobblestone walls
dev956

Yosshi999:
enable to place torches on cobblestone walls and glass

### DIFF
--- a/src/pocketmine/block/Torch.php
+++ b/src/pocketmine/block/Torch.php
@@ -55,7 +55,7 @@ class Torch extends Flowable{
 				0 => 0,
 			];
 
-			if($this->getSide($faces[$side])->isTransparent() === true and !($side === 0 and in_array($this->getSide(0)->getId(),array(self::FENCE,self::GLASS,self::STONE_WALL)))){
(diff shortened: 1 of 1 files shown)


src/pocketmine/block/Torch.php
```diff
@@ -78,7 +78,7 @@ public function place(Item $item, Block $block, Block $target, $face, $fx, $fy,
 			$this->getLevel()->setBlock($block, $this, true, true);
 
 			return true;
-		}elseif($this->getSide(0)->isTransparent() === false or $this->getSide(0)->getId() === self::FENCE){
+		}elseif($this->getSide(0)->isTransparent() === false or in_array($this->getSide(0)->getId(),array(self::FENCE,self::GLASS,self::STONE_WALL))){
```
PEMapModder: Suppose you should use `[]` instead?

Yosshi999: Do you mean
  array(self::FENCE, ...)
should be
  [self::FENCE, ...]
?

PR #2637, opened 2015-02-15
Allow array's in Player::sendMessage()

TruDan:
Means we can do Player::sendMessage(["Hi", "Hello Again!!", "You're Back!! Yay!"]);

Also reduces the need to have foreach($messages as $message) loops whenever sending multiple lines in one go.


shoghicp:
`Player->sendMessage(implode("\n", ["Hi", "Hello Again!!", "You're Back!! Yay!"]));`

Remember that Player extends CommandSender. That's where sendMessage() gets defined, and modifying that would break all the message handling.

Issue #2493, opened 2015-01-03
Chestbug


MobileCrafter:
Chests in other  direction can be placed against chests and are not transferring to a double chest.
(Sorry for bad english)
Also I think there are other issues with chests placement.

shoghicp:
Double chests are buggy, this is a known issue.

Creeperface01:
yes all items are despawning from double chest

wsantos:
I have the same issue

stxyzth:
Any fix on this yet?

prankard:
This was annoying me on my server. So I had a go at fixing it. Please test, check, test.

PR #2744, opened 2015-02-27
Added double chest fix

prankard:
Added a fix for the double chest Issue #2493

The data wasn't always going into the doublechest, and didn't appear to be combining correctly upon logout/login of server.

##### Double Chest Changes:

The doublechest, now starts with the items created from left/right in it's constructor.

Then in it's setContent's function. It now doesn't call parent (as the slot[] check is invalid and won't check both chests) but now has a similar pattern using the slots[] of left and right.

##### Minor improvement to base inventory:

When you set contents to an inventory, I now clear if the 'SetItem' fails.


shoghicp:
Could you follow the Contribution Guidelines? Specifically the code formatting, so it follows the project's formatting

PR #2751, opened 2015-03-02
Fixed bug where non-default generators are ignored in pocketmine\Server.php

alejandroliu:
Setting *FLAT* on `server.properties` or `pocketmine.yml` does not work, i.e. it is ignored by `LevelProvider::generator`.

The issue is that Server.php will create a new $generator instance, while `LevelProvide::generate` expects a string with the class name and an options array.

### DIFF
--- a/src/pocketmine/Server.php
+++ b/src/pocketmine/Server.php
@@ -1139,9 +1139,7 @@ class Server{
 
 		$seed = $seed === null ? Binary::readInt(@Utils::getRandomBytes(4, false)) : (int) $seed;
 
-		if($generator !== null and class_exists($generator) and is_subclass_of($generator, Generator::class)){
-			$generator = new $generator($options);
-		}else{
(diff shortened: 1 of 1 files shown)


alejandroliu:
The issue is that Server.php in line 1,143, it is creating a new $generator instance which is then used to call the selected LevelProvider::generate in line 1,156.  The signature for LevelProvider::generate is:

	 * Generate the needed files in the path given
	 *
	 * @param string  $path
	 * @param string  $name
	 * @param int     $seed
	 * @param string  $generator
	 * @param array[] $options
	 */
	public static function generate($path, $name, $seed, $generator, array $options = []);

So $generator shouldn't be a class instance but a string.  Indeed McRegion::generate does:

     "generatorName" => new String("generatorName", Generator::getGeneratorName($generator))

Which uses Generator::getGeneratorName which takes a string contain a full class name.

PR #2770, opened 2015-03-12
Fixed fallingsand bug

MinecrafterJPN:
Fixed bug
+ A chest will break when sand fall on the chest.

### DIFF
--- a/src/pocketmine/entity/FallingSand.php
+++ b/src/pocketmine/entity/FallingSand.php
@@ -108,7 +108,7 @@ class FallingSand extends Entity{
 			$this->motionY *= 1 - $this->drag;
 			$this->motionZ *= $friction;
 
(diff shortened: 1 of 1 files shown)


PEMapModder:
Maybe using `(int)` is better than using `floor()`.

Falkirks:
There is also a `round` method in `Vector3`.

PEMapModder:
@Falkirks 

X => floor
Y => round
Z => floor

```
G G G
   X
G G G

G G G
```
Suppose bottom left corner is x.0,z.0, `X` would be placed outside if we use `round()`.

For the Y, it is in case the block fell a bit more than the ground, then it would get sucked into the ground. Assuming that the block won't fall more than 0.5 block in a tick, rounding Y is more suitable.

MinecrafterJPN:
@PEMapModder Thank you

Issue #2741, opened 2015-02-25
If the server is a long operation, Corrupted chunk detected


HmHmmHm:
![1241241](https://cloud.githubusercontent.com/assets/8337568/6365116/f99e15c4-bcf4-11e4-8998-0a3e655f7c1e.png)

![14](https://cloud.githubusercontent.com/assets/8337568/6365150/8b1b8158-bcf5-11e4-91df-b568f8e1a775.png)
![17](https://cloud.githubusercontent.com/assets/8337568/6365149/8b1a3e1a-bcf5-11e4-982e-2c72ffef6ae6.png)

i'm using McRegion map and
PocketMine-Soft 238.

(it is 24 time server, always openning)

in the first boot everything is fine,
but after a lot of time (maybe 2day ~ 4day ago),
some chunk is overlapped generated ..

this error is extremely deadly...

how i can prevent or
stopping the generate overlap chunk?

![default](https://cloud.githubusercontent.com/assets/8337568/6365212/9c657f94-bcf6-11e4-88a2-b2be71e3979c.png)

64FF00:
Duudeeeee Try using ASR by @iJoshuaHD

HmHmmHm:
@64FF00 I aleady using for something my auto reload&reboot plug-in, (it's same ASR)

HmHmmHm:
aww.. i still have this problem..
why...why! corrupted chunk detecting??
![111111](https://cloud.githubusercontent.com/assets/8337568/6373431/f80251d6-bd4d-11e4-846d-546077e697ae.png)

iVertx:
I have the same issue:

![screenshot_2015-02-27-11-13-07](https://cloud.githubusercontent.com/assets/10237969/6424045/32a8b79e-becf-11e4-9b8b-298b571cb81e.png)
![screenshot_2015-02-27-12-50-50](https://cloud.githubusercontent.com/assets/10237969/6424051/40d63062-becf-11e4-83ff-c9e79741f8b3.png)

I did everything possible to solve this problem but nope... (Forget convert from Anvil to mcregion an vice-versa, this bug isn't related to it)
Man, it will simply destroy your map completelly like my map... And my map have other major bug like these = massive tree generations... I got tree walls in every grass block in my server, basically i can't build anything more in my map, because every, every and EVERY time i need to rebuild overlaped chunks and cut off trees... Really a dedly error in pocketmine

iVertx:
Captured this morning:

[...]

PR #2771, opened 2015-03-13
Fix some Chunk corruption due to Chunk overlap

alejandroliu:
When a chunks start going over 2 sectors, they begin to overlap which causes chunk corruption.

What happens is that `RegionLoader->lastSector` points to the last used sector.  When we try to save a Chunk that uses `2` sectors, the code afer line 191 gets executed.  The current behaviour is:

Extend lastSector by the number of sectors, then make the chunk index point to that lastSector.

This means that the chunk will start writing at the lastSector location but because it takes more than `1` sector of space, the next chunk that needs to allocate a sector will overwrite this chunk thus causing corruption.

This fix changes the code to:

Make chunk index point to lastSector+1, extent lastSector +  number of sectors needed.


alejandroliu:
I believe this fixes #2741.   The pictures show very elaborate builds which most likely spill to multiple sectors.

alejandroliu:
Please label PR:BugFix

shoghicp:
Thanks for the fix!

PR #2776, opened 2015-03-17
Implemented proportional armor modifier, applied armor in other damage types and consume armor when player is damaged

PEMapModder:


shoghicp:
Nice!
I'm going to review all of the pending pull requests today, so this might make it into 1.4.1 :)

src/pocketmine/Player.php
```diff
@@ -2685,7 +2659,23 @@ public function attack($damage, $source = EntityDamageEvent::CAUSE_MAGIC){
 		parent::attack($damage, $source);
 
 		if($source instanceof EntityDamageEvent and $source->isCancelled()){
```
PEMapModder: Ugh I forgot to delete `$source->isCancelled()`

shoghicp: Can you commit a patch? You can edit that file on your branch using the GitHub web interface, it'll show up here.

PEMapModder: is that another commit?

iJoshuaHD:
finally! now i dont have to hardcode this into my plugin :D

shoghicp:
I've reverted this, please open a new pull request when everything is fixed :)

Issue #2787, opened 2015-03-19
Chunk at 0,0 not loading in Flat worlds


alejandroliu:
Apparently Chunk at 0,0 does not load from file properly.

Can be modified and while in-memory seems OK and event gets saved correctly.  After unload, it doesn't get loaded and it gets re-generated.

alejandroliu:
To recreate this issue:

- Create a flat world.
- Place stuff at Chunk 0,0 (teleport to 8, 64, 8).
- Re-start server.
- The stuff at Chunk 0,0 is gone.

PEMapModder:
:+1:

alejandroliu:
This can be closed, as the fix in "Clone initial chunk in Flat.php #2788" has been merged.

PR #2788, opened 2015-03-19
Clone initial chunk in Flat.php

alejandroliu:
The Flat generator destroys Chunk at 0,0.

What happens is that it request the level to read Chunk at 0,0.  It then uses that chunk to create the template chunk.  However this obliterates whatever was in Chunk at 0,0.

Added a line to "clone" this chunk, so when parsePreset generates the template chunk, all this goes to a copy rather than the original file chunk 0,0.

This would fix issue #2787

### DIFF
--- a/src/pocketmine/level/generator/Flat.php
+++ b/src/pocketmine/level/generator/Flat.php
@@ -59,6 +59,7 @@ class Flat extends Generator{
 		$this->preset = "2;7,2x3,2;1;";
 		//$this->preset = "2;7,59x1,3x3,2;1;spawn(radius=10 block=89),decoration(treecount=80 grasscount=45)";
 		$this->options = $options;
+		$this->chunk = null;
 
 		if(isset($this->options["decoration"])){
 			$ores = new Ore();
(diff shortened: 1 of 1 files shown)


shoghicp:
I thought I had already fixed this... Still, you'll have an issue if that chunk contains something, it'll get duplicated over the world. The real solution would be calling the empty chunk method on the format, but you can't do that on another thread. I guess I'll move that to Chunk for 1.5

alejandroliu:
Yes, using "empty chunk" would be the right solution.

Let me take another go at it...

alejandroliu:
OK, second try.

I added code to delay the generation of the template chunk by `parsePreset` until we actually need to generate a block.

The idea is that when generateBlock is called, $chunkX and $chunkZ should always point to an empty Chunk.  We then parse presets and generate the template block from this $chunkX,$chunkZ (instead of the chunk at 0,0).  Then we start cloning this chunk as needed.

shoghicp:
This might not work if that chunk is unloaded and the owner is running the generator on the main thread. The chunk should be cloned as soon as it is received.

alejandroliu:
I added a "clone" so that the ownership of the chunk template is always the Flat generator.

I though that the situation that you describe doesn't happen because while parsePreset would read a new chunk from the level provider, and it is returned to generateChunk, generateChunk would then "clone" that chunk and return that to the level provider (thus replacing the original read chunk by parsePreset).

Essentially, this is doing what the "unpatched" Flat generator was doing but instead of hard coding the initial chunk to (0,0) it is getting that chunk from generateChunk.

PR #2794, opened 2015-03-21
Utils\Config preserves document type between reloads. Fixes #2777

### LINKED ISSUE #2777: Whitelist not working properly
When I add the first player my file contains this (I don't know why "type" is there):
```
type
test1
```
Added more:
```
type
test1
test2
```

Then I do a whitelist list and the output is:
```
White-listed players: type, test1, test2
```
After adding some more users and done some "whitelist list" and "whitelist reload" the file looks like this and users cannot login anymore.
```
{
    "type": true,
    "test1": true,
    "test2": true,
    "test3": true,
    "test4": true
}
```
The output of "whitelist list":
```
White-listed players: type, {, "type": true,, "test1": true,, "test2": true,, "test3": true,, "test4": true, }
```

sekjun9878:


sekjun9878:
@shoghicp Please confirm that this change will not affect any other parts of PocketMine and let me know

shoghicp:
I had already applied this patch to other branch and tested there, it should not affect things.

PR #2806, opened 2015-03-26
Fix #2735 Sand doesn't drop after explosion

### LINKED ISSUE #2735: Sand doesn't drop after explosion
Sand or gravel doesn't drop after explosion

Yosshi999:
Added BlockUpdateEvent

### DIFF
--- a/src/pocketmine/level/Explosion.php
+++ b/src/pocketmine/level/Explosion.php
@@ -51,6 +51,7 @@ class Explosion{
 	 * @var Block[]
 	 */
 	public $affectedBlocks = [];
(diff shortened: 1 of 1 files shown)


shoghicp:
Nice, I like this. Except that `in_array()` is O(n), making this O(n^2) at least. If you use the index as the key (as PocketMine does everywhere else) it'll be O(n)

Yosshi999:
So !in_array() should be !isset()?

sekjun9878:
Hello, I have test this patch however this patch only seems to affect the first (the lowest) block of sand, and not anything above it.

PR #2813, opened 2015-03-28
Really small fix

Gamecrafter:
Shouldn't the first reload message be the same as the second reload message? They should both be yellow.


sekjun9878:
Haha yes :)

PR #2882, opened 2015-04-14
Fixed colors not displaying on Windows

Flavius12:
Fixed colors not displaying on PM console on Windows users (using ANSI Escape Codes)
Hope I helped :)


shoghicp:
This breaks all the compatibility layers we implemented, please don't do this.

Flavius12:
Ah ok sorry

PR #2921, opened 2015-04-21
Zombie Modified AI

yungtechboy1:
A modified version of my previous pull request.
Now It is does not randomly locate a player.
It Locates the closest player.


shoghicp:
This still does unwanted side effects and does not follow the code format (see Contribution Guidelines).
Anyway, if an AI is implemented into PocketMine, it'll be a generic one that can be used for other entities and goals, not everything inside the same class.

Issue #2938, opened 2015-04-26
Players are invisible.


Vladik46:
Sometimes some players become invisible themselves are not visible but you can see how he puts the blocks.

P. S the server is locked, the command /effect!

PR #2942, opened 2015-04-26
Fixed issue #2938

Vladik46:


shoghicp:
This does not fix anything. If you follow the code execution, the spawn action has been already executed. Also, you are trying to spawn the player to itself, which won't work and if the checks were not there (for people that don't have any idea about how this works) you would have crashed all the players that join the server.

PR #2966, opened 2015-05-01
Is this still needed?

alejandroliu:
With MCPE0.10 very often when you teleport between worlds you will be looking at TileEntities (chests, signs, etc) from the originating world being left around the destination world.  You could see them, but you couldn't interact with them. (Couldn't touch or break them).

I am not part of the beta program, so I can not test this, but does this still happen in MCPE0.11?

If this still happens, I had code in ManyWorlds that would send UpdateBlockPackets before teleporting to clear-up those TileEntities so as to make sure that there are no tile entities that would get left behind after the new chunks are sent.  Would this fix that particular glitch?

Finally, if it does fix that particular glitch, would this patch be worth of consideration for merging into PocketMine-MP?

PS: Apologies, my text editor though it would be fun to remove the lines with only tabs with empty lines, so those changes also got to the commit.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -712,7 +712,7 @@ class Player extends Human implements CommandSender, InventoryHolder, IPlayer{
 			}
 
 			$this->noDamageTicks = 60;
-			
+
 			foreach($this->usedChunks as $index => $c){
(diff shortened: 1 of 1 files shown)


PEMapModder:
I think it is a duplicated issue.

PR #2969, opened 2015-05-01
Added Server->broadcastTip() and Server->broadcastPopup()

Striker209:


src/pocketmine/Server.php
```diff
@@ -1766,6 +1766,40 @@ public function broadcastMessage($message, $recipients = null){
 	}
 
 	/**
+	 * @param               $message
```
PEMapModder: Why no type?

TruDan:
:+1:

src/pocketmine/Server.php
Striker209MPE: I did the same as the Server->broadcastMessage() method but yes it should :D

sekjun9878:
Merged with adc5aa11ac26ae0668368f05dc01e19aae605b33. Thank you for the PR!

PR #2972, opened 2015-05-02
MuitlCore Support

ZhyMC:
I find that PM uses 2 of 16 cores to process tasks.
How can i use them.


shoghicp:
Please follow the Contribution Guidelines. If you continue to open spam-alike issues you may be blocked from this organization.

PR #2986, opened 2015-05-05
Fixed /tell command not displaying the sender's name correctly (aka "I'm Bored x3")

64FF00:
Ye, it really does.


64FF00:
No, it doesn't!

Issue #2999, opened 2015-05-10
PlayerDeathEvent Not Working


yungtechboy1:
I have Tried Everything! And I am almost 100% sure that PlayerDeathEvent() Dose not call! My Plugins that deal with Death are just spazing out!

PEMapModder:
Possible reasons:

* you didn't register the listener
* the player shows death screen but server doesn't think it dies. Client-side issue, won't fix.

yungtechboy1:
That the Thing I did register the Listener and It Does Show the Death Screen

PR #3001, opened 2015-05-10
Fixed invalid kill function

HmHmmHm:
I think it can solve the problem #2999


shoghicp:
If the death message is shown, PlayerDeathEvent is called. #2999 is just a client-side death (my guess in 0.10 and void)

HmHmmHm:
hmm.. anyway It is not just a problem PlayerDeathEvent
(plugins can not get PlayerDeathEvent in right time)

PR #3018, opened 2015-05-13
Push Commit for old Client

hmy2001:
In case client version is ~0.10.5, server doesnot send them a PlayerDisconnectPacket


shoghicp:
This is not needed as old clients won't accept this packet while not being logged in, it'll be ignored.

hmy2001:
@shoghicp ok.
I'm sorry.

PR #3032, opened 2015-05-18
Fix Crashes on Level.php

Samueljh1:
Fixes the Infamous getChunk() on null error!

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -2276,7 +2276,13 @@ class Level implements ChunkManager, Metadatable{
 
 		$this->cancelUnloadChunkRequest($x, $z);
 
-		$chunk = $this->provider->getChunk($x, $z, $generate);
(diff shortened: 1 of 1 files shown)


shoghicp:
Ehh...
No. This does not fix the issue and is a really bad hack.

Samueljh1:
Hack? And it does. On my SkyWars server, if I tried to unload and load a level, it can sometimes crash. I modified Level.php to check all usages of $this->provider is an instanceof err.. What was it? LevelBaseProvider? Anyway the server hasn't crashed since! (You should because the server will crashes when trying to load/unload some levels.) They are PC levels. It has happened a lot on my other server too!

shoghicp:
You might not be unloading worlds properly. Don't use Level->unload();

Samueljh1:
I used Server::getInstance->unloadLevel() should there be a different method?

shoghicp:
Also, if you are still getting chunks there are other issues with that level.

PR #3057, opened 2015-05-23
Added all tools to the creative inventory! (Build 13)

matcracker:
:+1:


matcracker:
Evidence: http://imgur.com/DJRlO5k

shoghicp:
We know, but we don't have the same items as in vanilla MCPE. Plugins can add their own items if they want.

matcracker:
Ok, sorry :)

PR #3058, opened 2015-05-23
Fix flowers backwards incompatibility issue.

sekjun9878:


shoghicp:
Constants are never removed, they are left for different makings. I'll check the other stuff later

PR #3066, opened 2015-05-25
Floor position parameters in Level::getBlock() and setBlock()

PEMapModder:


shoghicp:
No, plugins should do this before. Also, they usage of binary operators does that automatically. If any, this is fault of the caller. You can see that in PM any vector that might be a float is converted before calling it

PR #3080, opened 2015-05-29
Added Location::fromObject()

PEMapModder:


shoghicp:
This should be compatible with Position::fromObject().
You can put yaw and pitch at the end (after Level)

PEMapModder:
The argument list was created according to that of the constructor.

PEMapModder:
Closed due to missing use statement of `Vector3`

PR #3096, opened 2015-06-02
Changed if to switch

Striker209:
Looks better :D

### DIFF
--- a/src/pocketmine/entity/Item.php
+++ b/src/pocketmine/entity/Item.php
@@ -76,13 +76,13 @@ class Item extends Entity{
 	}
 
 	public function attack($damage, EntityDamageEvent $source){
-		if(
-			$source->getCause() === EntityDamageEvent::CAUSE_VOID or
-			$source->getCause() === EntityDamageEvent::CAUSE_FIRE_TICK or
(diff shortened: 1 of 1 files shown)


src/pocketmine/entity/Item.php
```diff
@@ -76,13 +76,13 @@ protected function initEntity(){
...
-			$source->getCause() === EntityDamageEvent::CAUSE_BLOCK_EXPLOSION
-		){
-			parent::attack($damage, $source);
+		switch($source->getCause()){
+			case EntityDamageEvent::CAUSE_VOID:
+			case EntityDamageEvent::CAUSE_FIRE_TICK:
+			case EntityDamageEvent::CAUSE_ENTITY_EXPLOSION:
+			case EntityDamageEvent::CAUSE_BLOCK_EXPLOSION:
+				parent::attack($damage, $source);
+				break;
```
LDX-MCPE: The `break` statement is unnecessary in this situation.

Humerus:
The `switch` statement is unnecessary in this situation.

sekjun9878:
The current implementation is not a problem. The if statement conveys intentions better and better looking code is subjective. Closing ticket as this is a non-issue.

PR #3129, opened 2015-06-07
Added  PluginBase::getResourceContents

alejandroliu:
Adds a convenient function to PluginBase called getResourceContents.

It is the same idea as file_get_contents, but for resources.

### DIFF
--- a/src/pocketmine/plugin/PluginBase.php
+++ b/src/pocketmine/plugin/PluginBase.php
@@ -187,6 +187,23 @@ abstract class PluginBase implements Plugin{
 	}
 
 	/**
+	 * Gets the contents of an embedded resource on the plugin file.
+	 *
+	 * @param string $filename
+	 *
+	 * @return string, or null
+	 */
+	public function getResourceContents($filename){
+		$fp = $this->getResource($filename);
+		if($fp === null){
+			return null;
+		}
+		$contents = stream_get_contents($fp);
+		fclose($fp);
+		return $contents;
+	}
+
+	/**
 	 * @param string $filename
 	 * @param bool   $replace
 	 *
--- a/src/pocketmine/plugin/PluginBase.php
+++ b/src/pocketmine/plugin/PluginBase.php
@@ -191,7 +191,7 @@ abstract class PluginBase implements Plugin{
 	 *
 	 * @param string $filename
 	 *
-	 * @return string, or null
+	 * @return string|null
 	 */
 	public function getResourceContents($filename){
 		$fp = $this->getResource($filename);


Falkirks:
:+1:

iksaku:
Good idea! :+1:

64FF00:
:+1:

src/pocketmine/plugin/PluginBase.php
```diff
@@ -187,6 +187,23 @@ public function getResource($filename){
 	}
 
 	/**
+	 * Gets the contents of an embedded resource on the plugin file.
+	 *
+	 * @param string $filename
+	 *
+	 * @return string, or null
```
sekjun9878: This is not a valid PHPDoc comment.

alejandroliu:
Sorry. Don't really now what's needed to make a valid PHPdoc.

src/pocketmine/plugin/PluginBase.php
PEMapModder: `string|null`
Refer to other parts of PocketMine code.

alejandroliu:
Sooo... took the wrong example from Line 78 (https://github.com/alejandroliu/PocketMine-MP/blob/PluginBase-suggestion/src/pocketmine/plugin/PluginBase.php#L178)

shoghicp:
Basically, this is a shortcut :)

In first place, this was intended to be a [SplFileObject](https://php.net/manual/en/class.splfileobject.php), but it lacks something very basic: ability to get the full contents. It also contains a lot of... interesting methods, that would fail or crash on phar archives.

alejandroliu:
So does that mean that you won't be merging this?

PEMapModder:
@shoghicp Then why not `PharFileInfo`? Its functions are quite interesting as well.

shoghicp:
@PEMapModder because that is dependent on the plugin being a phar, and that is only true in 1/3 of the possible plugin formats right now :)

alejandroliu:
@shoghicp, So in the meantime, can you merge this?

The other option to the @PEMapModder suggestion would be that the PluginLoader provides a factory to create the appropriate xxxFileInfo object.  So the PharLoader would return PharFileObject, the Folder loader would return SplFileInfo.  Any other loaders can return whatever is necessary.  The PHP snipped loader shouldn't throw and error as it wouldn't make sense for it.

shoghicp:
@alejandroliu PharFileInfo is just an SplFileInfo with the same issues as the original object - you can't get the full data directly. That's why that was not chosen.

PEMapModder:
Write your own class :stuck_out_tongue_closed_eyes:

alejandroliu:
I think just implementing the proposing getResourceContents from this PullRequest would take care of most use cases by most plugin developers.

sekjun9878:
I agree that a simple solution is more concise and accessible than trying to tie this into an object.

sekjun9878:
Doesn't look like this will cause any problems and it's BC compatible, so I'm merging this.

shoghicp:
This was not decided to be merged. I've force-pushed master back.

sekjun9878:
Why?

iksaku:
D:

PR #3146, opened 2015-06-09
Make it save to do logrorate at any time

alejandroliu:
This makes it possible to make it safe to do logrotate at any time.  What it does it keeps the file descriptor closed at all times and only opens it to do logging.


shoghicp:
We should not keep opening and closing file resources all the time, and the server (depending on the logger implementation that CAN be replaced by plugins) can do this or not, or print partial lines.
The only safe to do this is to stop, rotate, and start

PR #3156, opened 2015-06-10
Fix lingering tiles during world teleports

alejandroliu:
When somebody teleports between worlds Tile objects that used to be in the departure world keep lingering in the destination world.  These can be seen but cannot be interact with.
This pull request introduces a work around to solve these.

This is how the issue looks like:

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1638,7 +1638,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 
 					break;
 				}
-				
+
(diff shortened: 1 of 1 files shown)


shoghicp:
You might want to implement this on Player->switchLevel() instead of teleport

shoghicp:
Also, mark this code with `//TODO HACK` as I'll be looking for those later between version upgrades.

alejandroliu:
OK, but I am doing it as a different pull-request... that would be less commits (and easier for me than doing a rebase).

PR #3158, opened 2015-06-11
Fix lingering tiles during world teleports

alejandroliu:
New Pull Request in switchLevel

When somebody teleports between worlds Tile objects that used to be in the departure world keep lingering in the destination world.  These can be seen but cannot be interact with.
This pull request introduces a work around to solve these.

This is how the issue looks like:

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -590,6 +590,17 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	protected function switchLevel(Level $targetLevel){
 		$oldLevel = $this->level;
 		if(parent::switchLevel($targetLevel)){
+			//TODO HACK: removes tile entities that linger whenever you teleport
+			// to a different world
+			$pk = new UpdateBlockPacket();
+			foreach($oldLevel->getTiles() as $tile){
+				$pk->records[] = [$tile->x, $tile->z, $tile->y, 0, 0, UpdateBlockPacket::FLAG_NONE];
+			}
+			if(count($pk->records)){
+				$this->dataPacket($pk);
+			}
+			//----
+
 			foreach($this->usedChunks as $index => $d){
 				Level::getXZ($index, $X, $Z);
 				$this->unloadChunk($X, $Z, $oldLevel);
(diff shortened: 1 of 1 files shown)


matcracker:
:+1:

alejandroliu:
@shoghicp , is there anything else you need before this Pull Request can be merged?

AndreyNazarchuk:
Yay I was wondering ow to fix this :) :+1:

src/pocketmine/Player.php
```diff
@@ -590,6 +590,17 @@ public function isSleeping(){
 	protected function switchLevel(Level $targetLevel){
 		$oldLevel = $this->level;
 		if(parent::switchLevel($targetLevel)){
+			//TODO HACK: removes tile entities that linger whenever you teleport
+			// to a different world
+			$pk = new UpdateBlockPacket();
+			foreach($oldLevel->getTiles() as $tile){
+				$pk->records[] = [$tile->x, $tile->z, $tile->y, 0, 0, UpdateBlockPacket::FLAG_NONE];
+			}
+			if(count($pk->records)){
```
iksaku: ``count($k->records) > 0``

alejandroliu: It is enough to check that it is a non-zero value.

shoghicp:
@alejandroliu $oldLevel->getTiles() will send more data than needed, as that will get all from the entire world. What you want is to do that on unloadChunk() (so it's not implemented in two places) so you can get the Chunk tiles directly.

alejandroliu:
Will take a look at unloadChunk.

From what examining the code, $oldLevel()->getTiles() only sends the tiles that are loaded in memory.  Since chunks are dynamically loaded/unloaded, it should only send active tiles (which shouldn't be that many).

alejandroliu:
Oh, yes, the unloadChunk bit is much better... I will be updating this pull request.

shoghicp:
Can you squash this into a single commit so we don't have all the tries? Then force-push into your branch.

alejandroliu:
Created a new PR #3197 .  This one is a single commit.

PR #3175, opened 2015-06-15
Fixes #3173

### LINKED ISSUE #3173: Grass Growing Underwater
Using Dev 1.5 Build 1244, terrain generation adds grass underwater. If you go below a lake or sea, there is grass at the bottom. I don't think that is usual. Additionally, it is not grass growing down to dirt blocks underwater, for I found a grass-bottom lake in an immediate new chunk.

ExpandingDev:
Gets rid of useless Tall Grass Populator and fixes issue #3173 by setting the ground cover to just be dirt and not grass on top.

### DIFF
--- a/src/pocketmine/level/generator/normal/biome/OceanBiome.php
+++ b/src/pocketmine/level/generator/normal/biome/OceanBiome.php
@@ -28,10 +28,13 @@ class OceanBiome extends GrassyBiome{
 	public function __construct(){
 		parent::__construct();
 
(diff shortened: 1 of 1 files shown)


shoghicp:
This will happen in other places as well (rivers, a few biomes that go underwater)
This is not the way to fix it, as parts of the biome generation are not yet implemented, including underwater populators (with sand, gravel and dirt)

shoghicp:
Also this makes the build fail, this pull request is broken

PR #3182, opened 2015-06-17
Update Utils.php

SopaXorzTaker:
Correct a typo


SopaXorzTaker:
No need to push in master.

PEMapModder:
Where should it be if it isn't master?

SopaXorzTaker:
Oops, I have managed to mess it up. There is only master no develop.
Releases aren't made by branches.

SopaXorzTaker:
@PEMapModder I think it needs to be reopened.

PR #3183, opened 2015-06-17
Fix #3177

### LINKED ISSUE #3177: Invalid Local IP on Linux
When starting PocketMine, it determines the local IP as 127.0.0.1 instead of an actual one.

Tested on: Ubuntu 14.04 ARMv7 on ODROID-XU with latest dev version.

SopaXorzTaker:
Better gethostbyname local IP as per http://stackoverflow.com/questions/3219178/php-how-to-get-local-ip-of-system
Probably fixes #3177.


shoghicp:
"Probably" did you even test this?

PR #3184, opened 2015-06-17
Create Rail.php

RandomAltThing:
Another commit coming very soon to make it more functional

### DIFF
--- a/src/pocketmine/block/Rail.php
+++ b/src/pocketmine/block/Rail.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____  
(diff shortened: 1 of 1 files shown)


iksaku:
:-1: Rails are not shown to prevent people try to place "Minecarts", and they are not allowed in PocketMine because there are no physics for them yet :/

RandomAltThing:
Is there anyway to unblock rails in plugins? I managed to do it but it was SO glitchy. I don't want the physics yet, just the block.

PEMapModder:
If your pull request isn't functional yet, don't create it.

Adding the `Rails` class doesn't make rails become usable, and making rails place-able does not necessarily require adding the `Rails` class.

PR #3185, opened 2015-06-17
Prevent issues with colored display names

iksaku:
When plugins add custom "display names" with color codes, PocketMine and Chat messages are colored as the "display name" is... This is intended to correct that :3

For the Chat messages, the "suffix" of player's display name is colored too, for example, with the format ``<%s>%s``, the part ``%s>%s`` will be with the same color (Unless a chat plugin modify the format or the player use other color codes in chat)


TruDan:
:+1:

shoghicp:
Instead of fixing it on the call, this could be fixed on the parsing itself. That would be truly compatible with plugins, and would be dynamic if the color changes in the future.

iksaku:
Where do you suggest to place it?

PR #3197, opened 2015-06-19
Teleport glitch fix

alejandroliu:
Replaces PR #3158


alejandroliu:
OK, now it is in a single commit.

shoghicp:
D: there are extra tabs and spaces around the file

Beside that, seems ok, although only chests and signs need this. You can check if they are instances of `Spawnable` so it doesn't send extra data that is not needed.

alejandroliu:
Ugh... Normally I use emacs (yes, I am an old guy), and today was trying Atom.

Let me test the spawnable stuff...

PR #3205, opened 2015-06-20
removed extra breaklines :P

iJoshuaHD:


shoghicp:
¬¬
Why those two lines and not everything else?
Also, no

sekjun9878:
Yeah, those lines are there to separate between business logic and return value, but that double newline is weird.

PR #3220, opened 2015-06-24
Uniform /status format

sekjun9878:
Remove trailing dot for uniform /status command for easier parsing

### DIFF
--- a/src/pocketmine/command/defaults/StatusCommand.php
+++ b/src/pocketmine/command/defaults/StatusCommand.php
@@ -88,14 +88,14 @@ class StatusCommand extends VanillaCommand{
 
 		$sender->sendMessage(TextFormat::GOLD . "Thread count: " . TextFormat::RED . Utils::getThreadCount());
 
-		$sender->sendMessage(TextFormat::GOLD . "Main thread memory: " . TextFormat::RED . number_format(round(($mUsage[0] / 1024) / 1024, 2)) . " MB.");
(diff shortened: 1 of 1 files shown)


shoghicp:
Easier parsing, but you broke the existing parsing.

This was done to copy the format used in PC and other plugins (so existing code would be compatible with it), this now breaks compatibility with them completely.

sekjun9878:
Are you serious right now, this is starting to feel like Microsoft on their backward compatibility motto.. (oh wait! :P)

Examples that this format is used in PC and will result in serious stuff braking?

shoghicp:
Anyway, this does not make parsing "easier", it's the same, just an extra character.
`/^([^:]+): ([^\.]+)\.$/` with dot
`/^([^:]+): (.+)$/` without dot

then remove the special chars and you have everything split by key and value

sekjun9878:
*sigh* Why would you even make a commit with MBs with dots appended and Kb/s without a dot appended? This is a fix, not "breaking backwards compatibility"... I don't understand how an issue as critical as an unnecessary dot being appended taking up 4 extra bytes doesn't come across as an issue that requires IMMEDIATE fixing?!

It literally BURNS my eyes every time I run the status command, and I weep at the lack of uniformity between the lines how can you not see this as a problem it needs to be FIXED NOW!!!111

sekjun9878:
Whoops, I seem to have dropped this -> /s

sekjun9878:
@shoghicp Any update on this? It makes sense to drop the period.

PR #3221, opened 2015-06-24
Allow player to send message to console

RandomAltThing:


shoghicp:
This adds extra stuff that is not specified, and does not follow the syntax, and is not compatible with plugins.
---
Remember that you were blocked from this repository, if you continue doing this, you will be reported to the GitHub staff again.

PR #3236, opened 2015-06-28
fix #3232 dropping only cobblestone

### LINKED ISSUE #3232: Andesite, Granite and Diorite drop cobblestone.

Yosshi999:


src/pocketmine/block/Stone.php
```diff
@@ -77,12 +77,18 @@ public function getBreakTime(Item $item){
 
 	public function getDrops(Item $item){
 		if($item->isPickaxe() >= Tool::TIER_WOODEN){
-			return [
-				[Item::COBBLESTONE, 0, 1],
```
PEMapModder: Why not simply replace this line into `return [[Item::COBBLESTONE, $this->meta, 1]];`?

sekjun9878: Personally I wouldn't put two array brackets like that - it's not obvious that the array is supposed to be a list.

shoghicp:
This is being changed in new-blocks

src/pocketmine/block/Stone.php
Yosshi999: >>PEMapModder
Because the id of andesite etc is 1, not 4.(4 is cobblestone's id)

PEMapModder: @Yosshi999 from what I see in your code implementation, there is no difference.

PR #3245, opened 2015-07-01
Fixes lastPlayed

alejandroliu:
LastPlayed value is not being saved correctly.  Right now saves it in a tag that is the same as the timestamp (so player files just keep growing with lots of timestamp tags).


PEMapModder:
:+1:

PR #3247, opened 2015-07-02
Fixes first respawn when spawning in non-default level.

TruDan:
Does as it says in the title. Fixes when the very first PlayerRespawnEvent sets a position in a different level.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -746,6 +746,8 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 		$pk->z = $pos->z;
 		$this->dataPacket($pk->setChannel(Network::CHANNEL_WORLD_CHUNKS));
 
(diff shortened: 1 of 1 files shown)


shoghicp:
You commented that this did not fix your issue on IRC, right?

TruDan:
Yea, i tested with this commit and the issue was that chunks are sent before this is even called. So it makes it very complicated to change the level for their first respawn event. 

I worked around it by setting the level in PlayerLoginEvent, and then setting the position in playerrespawnevent.

shoghicp:
I read this again, but you can only set the level at login / pre login via configuration. Setting it on Join is a **really bad idea** as players already got half of the other world sent to them.

Even if this worked, you are causing lots of issues in clients on other parts of the code, change the level before the player starts doing things.

TruDan:
Yeah, even then it kinda works, i think i did some wierd hacks and override the saved data to return a different world

PR #3287, opened 2015-07-19
Added all stone blocks drops when breaking different stones

Flavius12:
Now when breaking these stone blocks it won't drop the default cobblestone but the right block:
- Granite
- Polished Granite
- Diorite
- Polished Diorite
- Andesite
- Polished Andesite

Hope I helped you :)


shoghicp:
This is already implemented in the correct branch.

Also you should learn about if statements and conditional data

PR #3296, opened 2015-07-21
clean formatting mistakes

vvzar:


PEMapModder:
http://php.net/manual/en/language.references.pass.php
This is called `parameter passed by reference`. The original code is correct, and your code breaks the functionality of the function.

jython234:
Kind of like C pointers.

sekjun9878:
Actually

PR #3331, opened 2015-07-30
Possible workaround with #3330

### LINKED ISSUE #3330: PluginManager attempts to register non event handling methods
When `PluginManager->registerEvents(Listener, Plugin)` is called, all methods in `listener` with a parameter extending `Event` will be registered as event handlers, which could bring unexpected results. 

For example:

```php
public function getDeathMessage(EntityDamageEvent $event) {
	$message = TextFormat::YELLOW.TextFormat::BOLD;
	switch ($event - > getCause()) {
		case EntityDamageEvent::CAUSE_CONTACT:
			$message. = 'you were killed by a cactus';
			break;
		case EntityDamageEvent::CAUSE_SUFFOCATION:
			$message. = 'you suffocated to death';
			break;
			// etc.
	}
	return $message;
}
```

Consider this method is included in a `Match` class that represents a minigame match, which also contains event handlers for events such as `PlayerDeathEvent` or `EntityDamageByEntityEvent`. Since the class would have to register events by calling `PluginManager->registerEvents(Listener, Plugin)`, the above method would also be treated as an event handler, while it is not.

The worse case:

```php
protected function getDeathMessage(EntityDamageEvent $event) // Changing modifier from public to protected
{
	//...
}
```

Even though a method is protected, `PluginManager` still registers it as an event handler, and would call the method when `EntityDamageEvent` is fired, hence crashing the server:

```
Fatal error: Call to protected method ...getDeathMessage() from context 'pocketmine\plugin\MethodEventExecutor'
```

My suggestion: either adding a PHPdoc annotation just like `@priority` and `@ignoreCancelled`, like `@nonEventHandler`:

```php
/**
 * @nonEventHandler
 */
public function getDeathMessage(EntityDamageEvent $event) // This method would not be treated as an event handler
{
    //...
}
```

Or, standardize `@EventHandler` just like Bukkit.

PEMapModder:
Filters away non-public event handlers, fixes #3330

### DIFF
--- a/src/pocketmine/plugin/PluginManager.php
+++ b/src/pocketmine/plugin/PluginManager.php
@@ -701,7 +701,7 @@ class PluginManager{
 		}
 
 		$reflection = new \ReflectionClass(get_class($listener));
-		foreach($reflection->getMethods() as $method){
+		foreach($reflection->getMethods(\ReflectionMethod::IS_PUBLIC) as $method){
 			if(!$method->isStatic()){
 				$priority = EventPriority::NORMAL;
 				$ignoreCancelled = false;


sekjun9878:
Sounds good. This won't break backwards compatibility because private and protected methods can't be called anyway.

PR #3332, opened 2015-07-30
Disallow registration of events without handlerList

PEMapModder:
Fixes another issue mentioned in #3330

### DIFF
--- a/src/pocketmine/plugin/PluginManager.php
+++ b/src/pocketmine/plugin/PluginManager.php
@@ -748,7 +748,8 @@ class PluginManager{
 	 * @throws PluginException
 	 */
 	public function registerEvent($event, Listener $listener, $priority, EventExecutor $executor, Plugin $plugin, $ignoreCancelled = false){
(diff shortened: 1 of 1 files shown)


TigerHix:
Good one. It is quite common to register events without `handlerList`, such as `EntityDamageByEntityEvent` which extends `EntityDamageEvent`. Strangely and confusing though, `PlayerDeathEvent` that extends `EntityDeathEvent` has the `handlerList` field.

sekjun9878:
Could you give me an example of what this is supposed to fix? #3330 doesn't give any examples and I don't deal with PM enough to know what problems this causes.

sekjun9878:
Just to clarify, this is a purely syntactic sugar change. This is estimated to break some plugins that incorrectly utilise some inherited Events, but that is an incorrect usage and not covered by the BC guarantee. 

This check makes sure that there are no type hint errors by making it impossible to Events that are a child of another event, but don't have its own $handlerList. 

@PEMapModder Add a nice little message and I'll merge this right away

sekjun9878:
It's been merged!

PR #3345, opened 2015-08-05
Update Human.php

PeratX:
Fix crash when player join and quit


shoghicp:
Actually it's clientId instead of clientID on remove packet. I'll just push something else, as we already have two pull requests that don't fix it completely.

PR #3359, opened 2015-08-09
Mcpe 0.12

vellrya:
Freezing, during destruction or installation of the first block. Also it is impossible to take the block in a creative. Fix please)


sekjun9878:
Merging between branches of the PocketMine-MP Repository should only be handled by PocketMine Git Team Members. Closing Pull Request.

PR #3360, opened 2015-08-09
$handlerList existence check upon event handler registration, closes #3332

PEMapModder:


PEMapModder:
Oops, forgot to define $class

PR #3385, opened 2015-08-17
Prevent plugins crashing the server by requesting permissions for users that have gone offline

mwvent:
This is of course an issue that should be fixed plugin side but without this defensive coding it is hard to even identify what plugin is causing the issue - the error I was getting constantly before in the crash logs on our server was

PocketMine-MP Crash Dump Sat Aug 15 09:56:24 BST 2015                                                                                [0/624]

Error: Call to a member function hasPermission() on null
/src/pocketmine/Player__64bit
Line: 439
Type: E_ERROR

Code:
[430]           return $this->perm->isPermissionSet($name);
[431]   }
[432]
[433]   /**
[434]    * @param permission\Permission|string $name
[435]    *
[436]    * @return bool
[437]    */
[438]   public function hasPermission($name){
[439]           return $this->perm->hasPermission($name);
[440]   }
[441] 
[442]   /**
[443]    * @param Plugin $plugin
[444]    * @param string $name
[445]    * @param bool   $value
[446]    *
[447]    * @return permission\PermissionAttachment
[448]    */
[449]   public function addAttachment(Plugin $plugin, $name = \null, $value = \null){

Backtrace: 
#0 (): pocketmine\Server->crashDump()

Now after running with this change I can see many prevented crashes

cat server.log | grep "assuming false"
2015-08-17 [02:01:22] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [02:01:22] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [02:01:22] [Server thread/WARNING]: Failed to read permission essentials.afk.kickexempt because Player has gone - assuming false
2015-08-17 [02:19:22] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [02:19:22] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [02:19:22] [Server thread/WARNING]: Failed to read permission essentials.afk.kickexempt because Player has gone - assuming false
2015-08-17 [03:43:52] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [03:43:52] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [04:32:22] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [04:32:22] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [05:05:52] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [05:05:52] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [08:31:23] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [08:31:23] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [10:26:24] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [10:26:24] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [10:36:54] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [10:36:54] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [10:36:54] [Server thread/WARNING]: Failed to read permission essentials.afk.kickexempt because Player has gone - assuming false
2015-08-17 [11:13:25] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [11:13:25] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [17:16:41] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [17:16:41] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [17:51:19] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [17:51:19] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [18:29:50] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [18:29:50] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [19:06:20] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [19:06:20] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [19:06:20] [Server thread/WARNING]: Failed to read permission essentials.afk.kickexempt because Player has gone - assuming false
2015-08-17 [21:39:57] [Server thread/WARNING]: Failed to read permission essentals.afk.preventauto because Player has gone - assuming false
2015-08-17 [21:39:57] [Server thread/WARNING]: Failed to read permission essentials.mute.notify because Player has gone - assuming false
2015-08-17 [21:39:57] [Server thread/WARNING]: Failed to read permission essentials.afk.kickexempt because Player has gone - assuming false

So I can at least see that Essentials is causing the issue now and the issue is no longer fatal.

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -436,6 +436,14 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 	 * @return bool
 	 */
 	public function hasPermission($name){
+		if(is_null($this)) {
+			$this->server->getLogger()->warning("Failed to read permission " . $name ." because Player has gone - assuming false");
+			return false;
+		}
+		if(is_null($this->perm)) {
+                        $this->server->getLogger()->warning("Failed to read permission " . $name ." because Player has gone - assuming false");
+                        return false;
+                }
 		return $this->perm->hasPermission($name);
 	}
 


shoghicp:
Oh no, a warning would be worse. Please request this to be fixed on the plugins themselves, not here (as a player that has been disconnected is on an invalid state, doing stuff with it can crash OR leave the server in an unusable state (even corrupt everything).

The decision to leave it like that was made from the start.

mwvent:
Ok duly noted. Thank you :+1:

PEMapModder:
What? `if(is_null($this))`?

mwvent:
It does return true occasionally in our case - its saved a lot of crashes on our server having the null check. I will report to the EssentialsPE author when I get a chance and for now Ill keep this band-aid fix on my server with regular backups so we can rollback if we do experience any corruption as shoghicp mentioned . And thanks to you guys for making this software it makes a lot of people so happy :-)

PEMapModder:
@mwvent if it is null, there isn't an object at all. If it isn't an object, you won't have the chance to call from its context.
Also, we are fixing the issue at the @LegendOfMCPE team.

PEMapModder:
@shoghicp I just looked at the code a bit, and realized that the `hasPermission` method and relevant methods come from the `Permissible` interface. However, it appears that if you hold a `Permissible` object (or more practically, a `CommandSender` object), you are unable to check whether the object is still permissible (permission operations can still be operated upon it), unless you explicitly check `!($permissible instanceof Player) or $permissible->isOnline()`, which does not seem to be the best thing that can appear in code. I am suggesting that if you are doing this, you'd better also add a function like `isOnline` or `isAvailable` at the `Permissible` interface.

I am working on a plugin that allows command senders to instantiate a new `CommandSender`, whose permissions inherit the ones from the parent command sender. The implementation is something like this:

```php
class CommandControlledUser implements CommandSender{
	/** @var CommandSender */
	private $parent; // I really want to use WeakRef for this :(

	public function isPermissionSet($name){
		return $this->parent->isPermissionSet($name);
	}
	public function hasPermission($name){
		return $this->parent->hasPermission($name);
	}
	public function addAttachment(Plugin $plugin, $name = null, $value = null){
		return $this->parent->addAttachment($plugin, $name, $value);
	}
	public function removeAttachment(PermissionAttachment $attachment){
		$this->parent->removeAttachment($attachment);
	}
	public function recalculatePermissions(){
		$this->parent->recalculatePermissions();
	}
	public function getEffectivePermissions(){
		return $this->parent->getEffectivePermissions();
	}
	public function isOp(){
		return $this->parent->isOp();
	}
	public function setOp($value){
		$this->parent->setOp($value);
	}
}
```

I had thought of making $this->parent a WeakRef, but I had heard that this extension is not supported in future releases. I had also thought of garbage-collecting these command-controlled users upon player disconnection, but what about non-player command senders? For example, if someone makes an IRC bridge (that starts up an IRC **server**, _not client_), he may create a class `IRCComamndSender implements CommandSender`, and when the IRC command sender executes a command that creates a `CommandControlledUser` for them, I don't think it is possible to know about the disconnection of the IRC command sender, and when someone attempts to do something upon this obsolete command-controlled user from the IRC command sender, such as from a scheduled task, there is no way to check whether the IRC user is still permissible.

===

Of course, other problems arise from a similar situation too. For instance, the command-controlled user will not be collected, and eventually results in memory leak. This is an issue with the `CommandSender` API itself.

PR #3392, opened 2015-08-20
Handle script plugin lists

alejandroliu:
The plugin.yml has some items that are supposed to be arrays.  This changes the ScriptPluginLoader to make sure that those items are indeed arrays.

### DIFF
--- a/src/pocketmine/plugin/ScriptPluginLoader.php
+++ b/src/pocketmine/plugin/ScriptPluginLoader.php
@@ -106,6 +106,15 @@ class ScriptPluginLoader implements PluginLoader{
 			}
 		}
 		if($insideHeader){
+			/*
+			 * We go through the description and make sure that array
+			 * tags are indeed arrays...
+			 */
+			foreach(["depend", "softdepend", "loadbefore", "api", "authors"] as $key){
+				if(isset($data[$key])) {
+					$data[$key] = preg_split("/\s*,\s*/", $data[$key]);
+				}
+			}
 			return new PluginDescription($data);
 		}
 


shoghicp:
They were kept like that for a reason - scripts are only intended to be fast ways to interact with the core and experiment, not create full plugins. Also adding special cases for arrays does not seem like a good option :(

Also this does not handle keys!

alejandroliu:
While scripts are only intended to be fast ways to interact with the core and experiment, sometimes is useful to control the order of plugins through depend and softdepend.

alejandroliu:
BTW, this patch is not trying to handle all cases (i.e commands and permissions).  It simply addresses the plugin metadata elements that are just lists (no keys) and usually are simply casted to array.

Normally in plugin.yml these fields are entered as:

key: [ value,values]

alejandroliu:
Also, if we wanted to handle keys, I think I can re-implement it so that tags are extracted and converted to YAML and then the YAML parser is called.   But I do think that would be over the top.

shoghicp:
Can you generalize this patch so it supports any tag with array and key support?
Otherwise adding special cases is not nice.

alejandroliu:
As "special cases" go, this one targets the special cases already being defined in "PluginDescription".  Anyway, perhaps, the "special case handling" should be moved to PluginDescription (if you prefer).
Either way, I included two additional Pull request that are more generic.   One only implements arrays, and the other one implements array and key support.  You obviously only need to pick one from these three pull requests.

PrimusLV:
Not always key 'api' should be array!

```yaml
api: 1.0.0 # Will work for all API versions
api: [1.0.0] # Will it now?
```

PEMapModder:
PluginDescription would automatically cast it to an array.

alejandroliu:
bitrot

PR #3397, opened 2015-08-22
Use YAML to parse ScripPlugin headers

alejandroliu:
Uses "yaml_parse" to parse ScriptPlugin headers.  With that ScriptPlugin can support all the features that a normal plugin.yml can use.  The downside is that it is subject to YAML indentation idiosyncrasies.  

NOTE: I am not sure what is your preference, so I created several pull-request so you can pick and choose what suits you better.


alejandroliu:
bitrot

PR #3398, opened 2015-08-22
Implemented basic lists

alejandroliu:
Extends the ScriptPlugin metadata parses so it understands the YAML like notation of using "[" and "]" to denote arrays.  So you could use:

     @authors [ alice, bob, carlos ]

NOTE: I am not sure what is your preference, so I created several pull-request so you can pick and choose what suits you better.

### DIFF
--- a/src/pocketmine/item/CookedFish.php
+++ b/src/pocketmine/item/CookedFish.php
@@ -24,7 +24,7 @@ namespace pocketmine\item;
 
 class CookedFish extends Item{
 	public function __construct($meta = 0, $count = 1){
-		parent::__construct(self::RAW_FISH, $meta, $count, "Cooked Fish");
+		parent::__construct(self::COOKED_FISH, $meta, $count, "Cooked Fish");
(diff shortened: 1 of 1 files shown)


src/pocketmine/plugin/ScriptPluginLoader.php
```diff
@@ -97,7 +97,10 @@ public function getPluginDescription($file){
 			if(preg_match("/^[ \t]+\\*[ \t]+@([a-zA-Z]+)[ \t]+(.*)$/", $line, $matches) > 0){
 				$key = $matches[1];
 				$content = trim($matches[2]);
-
+				// Check if the user intended this to be an array...
+				if($content{0} == "[" and substr($content,-1)){
```
shoghicp: I think you forgot something here: `substr($content,-1)`, wouldn't that be `substr($content,-1) === "]"` (or `$content{strlen($content) - 1} === "]"`)?

Also please use `===` instead of `==` in the first condition.

PR #3399, opened 2015-08-22
Improve string to array conversions

alejandroliu:
Another way to handle the "special cases".  fields that should be arrays are parsed (through __toArray) so commas are treated as separators.

Not really needed for plugin.yml (which uses Yaml parse) if the developer did things properly.  However, if things somehow were not done properly, this makes sure PluginDescription handles them properly.

For ScriptPlugins that can only pass strings for all fields, this makes sure that depend and softdepend get reasonable values.

### DIFF
--- a/src/pocketmine/plugin/PluginDescription.php
+++ b/src/pocketmine/plugin/PluginDescription.php
@@ -45,6 +45,17 @@ class PluginDescription{
 	private $permissions = [];
 
 	/**
+	 * @param string|array $str
+	 */
+	static private function forceArray($str){
+		if(is_array($str)){
+			return $str;
+		}
+		return preg_split("/\\s*,\\s*/", $str);
+	}
+
+
+	/**
 	 * @param string|array $yamlString
 	 */
 	public function __construct($yamlString){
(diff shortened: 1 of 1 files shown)


src/pocketmine/plugin/PluginDescription.php
```diff
@@ -45,6 +45,17 @@ class PluginDescription{
 	private $permissions = [];
 
 	/**
+	 * @param string|array $str
+	 */
+	static protected function __toArray($str){
```
PEMapModder: Why `__`?

Falkirks: Looks like `__toString`, but isn't magical, so a bit of a misnomer.

alejandroliu: Yes, didn't know what to name the function.  Can anybody suggest a better name?
It is meant to make sure that arrays are arrays (or that strings are converted to arrays).

PEMapModder: Maybe just `private static function forceArray()`?

alejandroliu:
Yes, that probably would be much better.

I will re-do this PR when I get home.

alejandroliu:
bitrot

PR #3414, opened 2015-08-25
fixed item can not be seen in the inventory after give

Creeperface01:


SkygridRex:
Like

PEMapModder:
`addItem` is supposed to send the item. It isn't proper to fix the bug without fixing the API function addItem itself.

PR #3422, opened 2015-08-27
Update SayCommand.php

ondraczsk:
Update to czech language. Sorry IM reupdate after 20 hours. My english speak is horrible


willowmaster66:
I think if you do this everyone will see the new text. Even if you are not using the czech language.

Take a look at the Server->getLanguage()->translateString function and the files in pocketmine\lang\locale. There is also a README.md in this folder you should check out.

SkygridRex:
Its for 0.12 path?

legoboy0215:
:+1:

PR #3423, opened 2015-08-27
Block cache not cleared with a call to clearcache

willowmaster66:
While trying to figure out Github and looking for a different bug I found this small bug in the code. The ClearCache function on Level.php did not correctly clear the blockcache.


SkygridRex:
But, can you add chunckcache too?

willowmaster66:
It is already there. Only the block cache did not work because it cleared the chunck cache. It was a copy and past error. This will only be a "problem" if it is not a forced cache clear.

PR #3424, opened 2015-08-27
Redstone has to break 5 times to be finally mined.

willowmaster66:
You had to break the redstone 5 times before it was finaly broken.

### DIFF
--- a/src/pocketmine/block/GlowingRedstoneOre.php
+++ b/src/pocketmine/block/GlowingRedstoneOre.php
@@ -24,8 +24,9 @@ namespace pocketmine\block;
 use pocketmine\item\Item;
 use pocketmine\item\Tool;
 use pocketmine\level\Level;
+use pocketmine\block\RedstoneOre;
 
-class GlowingRedstoneOre extends Solid{
+class GlowingRedstoneOre extends RedstoneOre{
 
 	protected $id = self::GLOWING_REDSTONE_ORE;
 
@@ -33,16 +34,8 @@ class GlowingRedstoneOre extends Solid{
 
 	}
 
-	public function getHardness(){
-		return 15;
-	}
-
 	public function getName(){
 		return "Glowing Redstone Ore";
-	}
-
-	public function getToolType(){
-		return Tool::TYPE_PICKAXE;
 	}
 
 	public function getLightLevel(){
@@ -57,16 +50,6 @@ class GlowingRedstoneOre extends Solid{
 		}
 
 		return false;
-	}
-
-	public function getDrops(Item $item){
-		if($item->isPickaxe() >= 4){
-			return [
-				[Item::REDSTONE_DUST, 0, mt_rand(4, 5)],
-			];
-		}else{
-			return [];
-		}
 	}
 
 }


legoboy0215:
Why delete those? This is totally random..

willowmaster66:
It is not random. I change the extend from Solid to the Redstone class. The same functions are already in the Redstone class.

SkygridRex:
Delete line 55

PEMapModder:
I don't see any reason this pull request fixed anything

legoboy0215:
:+1:

sekjun9878:
The change is unintuitive - surely there's a simpler solution. It could be something to do with the hardness setting and that is a single line change.

willowmaster66:
I don't agree with your point of view. What I did is called Object Oriented Programming. If OOP had been used in the first place it would not have led to this bug. It is only unintuitive if you are not familiar with OOP. I will make a new pull request to only fix the hardness so this bug is fixed, but please let shoghicp review this pull request because I think he will understand. He has made some many good OOP decisions.

legoboy0215:
Umm... Maybe PEMapModder does not know OOP??? *Half-Sarcasm*

willowmaster66:
I don't know the work of PEMapModder or if he understands OOP nor do I want to say someone is stupid. I'm a software engineer for 8 years now and even I haven't grasped all the aspects of OOP. It is not easy to do it right. You need to be able to think abstract at a high level. Even I fail at this many times. I just want to contribute not set any bad blood. I have the best intentions and apologize if I have offended anyone.

legoboy0215:
lol You did not offend us. It is up to them to decide :8ball:

PEMapModder:
@legoboy0215 please note that this is not a forum.
@willowmaster66 but that doesn't seem necessary. You could just change the hardness of the block.

legoboy0215:
Ok. :sob:

willowmaster66:
I agree that at first that seems the logical thing to do, but if you compare the RestoneOre.php with the GlowingRedstoneOre.php code you see a lot of the code is exactly the same. And that is bad programming practice. Duplicate code means it's bug prone.Some day someone decides to make it harder to mine the redstone and changes the RedstoneOre.php accordingly, but (as happend here) does not know about the GlowingRedstoneOre.php You will have the same problem again. By extending the GlowingRedstoneOre class with the RedstoneOre class you automatically get all the code from the RedstoneOre. You simply only add extra functionality or override the functionality you think is different. As with the OnUpdate function. The hardness has to be the same so you can remove it here. 

Think of it like this. You have a class Person with properties Eyecolor and Haircolor. It also has a function onDanger. If onDanger is called the automatic reaction of the person is to run away which is the function RunAway. Then you decide to implement a Villager class and a Hero class, because they act differently on a dangerous situation. So you create a Villager class and a Hero class that extend the Person class. You get the haircolor and eyecolor for free so don't have to write those again. Then for the Hero class you override the onDanger function, because the hero will fight and not run away. So you create the onDanger function in the hero class that overrides the behaviour of the person class. This time it calls a newly created function Fight. When the fighting does not go well the Hero could eventually call the RunAway fucntion so he will not be killed.

If I had not created a Person class but directly the Hero and Villager class I could have forgotten about implementing the haircolor or eyecolor and there would not have been a default behaviour.

Please if you need me to explain this even better just ask. I'm happy to try to make this understandable for everyone which is quit a challenge for me (and something I can learn from :) ). I could make a code example if you want me to.

PEMapModder:
:+1:

sekjun9878:
While you do make some good points, confusing business-logic-level inheritance to mean code should use inheritance, and/or over using inheritance are common beginner mistakes that make code vulnerable to the "fragile base class" anti-pattern.

That's not the sole reason I rejected this (this change is needless) but I just wanted to let you know that inheritance is not always the right tool to use.

SkygridRex:
@willowmaster66 for me this change it's ok
Duplicate code it's a bad practice because can be make conflict

willowmaster66:
@sekjun9878 I love patterns and anti-patterns. I'll look it up. Good to know there are people running this project that actually do know what they are doing.

@SkygridRex At first I thought too, but I didn't know about the anti-pattern. Have to check it out first.

sekjun9878:
@willowmaster66 FYI that Inheritance in code should be based on something solid and specifically designed to be inherited from, *not because it relates to real life*.

Also, less layers of inheritance is a good thing. The less the better. After a few layers it just becomes a monstrosity of "I don't know what part of the inheritance-chain this goes in"

I don't know if you already know this or not, but just wanted to rant :D

willowmaster66:
Keep on ranting. :) I have read the fragile base class anti pattern and it's actually quite logical.

PR #3426, opened 2015-08-29
Remove unused parameter in Level::requestChunk

PEMapModder:


sekjun9878:
Sounds good. PHP doesn't check user-defined function parameters strictly so this should be BC compatible.

SkygridRex:
I see preference for some users (like a PeMapModder)

PEMapModder:
@SkygridRex please note that this is an issue tracker, not a forum.

sekjun9878:
@SkygridRex Not really... Usually I just see some new PR on my GitHub feed, and respond to them so.. I'll look at the others when I have the chance.

shoghicp:
This was an old parameter that was not removed waiting for the API change... But I guess it's ok to do it now

PR #3427, opened 2015-08-29
Removed redundant parameters for Living::knockBack

PEMapModder:
For valid calls on this function, it should be backwards-compatible.

### DIFF
--- a/src/pocketmine/entity/Living.php
+++ b/src/pocketmine/entity/Living.php
@@ -132,7 +132,18 @@ abstract class Living extends Entity implements Damageable{
 		$this->attackTime = 10; //0.5 seconds cooldown
 	}
 
-	public function knockBack(Entity $attacker, $damage, $x, $z, $base = 0.4){
+	/**
+	 * WARNING: Do not pass more arguments to this function than needed.
+	 * @param number $x
+	 * @param number $z
+	 * @param float  $base default 0.4
+	 */
+	public function knockBack($x, $z, $base = 0.4){
+		if(func_num_args() >= 4){
+			$args = func_get_args();
+			list(, , $x, $z) = $args;
+			$base = isset($args[4]) ? $args[4] : 0.4;
+		}
 		$f = sqrt($x * $x + $z * $z);
 		if($f <= 0){
 			return;


sekjun9878:
And how does this help exactly?

src/pocketmine/entity/Living.php
```diff
@@ -132,7 +132,18 @@ public function attack($damage, EntityDamageEvent $source){
...
 	}
 
-	public function knockBack(Entity $attacker, $damage, $x, $z, $base = 0.4){
+	/**
+	 * WARNING: Do not pass more arguments to this function than needed.
+	 * @param number $x
+	 * @param number $z
+	 * @param float  $base default 0.4
+	 */
+	public function knockBack($x, $z, $base = 0.4){
```
sekjun9878: Note: This really is BC compatible because of the func_get_args below. The warning message should be removed, however. Rather put a comment on the BC part saying why that exists there for future maintainers.

PEMapModder:
Removes redundant parameters? It would be really pointless to keep telling developers to find an entity object (although it can be self) just for piling up the parameter count.

shoghicp:
Why was this merged? Entity is there because some entities might want to know the entity. If any, it should be able to be set to null (make it nullable with ` = null`), not change the signature.

This also breaks compatibility with existing classes that extend this method.

@sekjun9878 Please revert this now

PEMapModder:
@shoghicp any changes with the signature still would break BC. Even a null default or a type hint change would.

sekjun9878:
@shoghicp There's no class that overrides the knockback function so I don't see what you mean by where "some entities want to know the entity" - the parameter isn't being used anywhere. 

Furthermore, PHP doesn't have strict parameter checking - child classes override not overload so there should be no BC breaks (as in, if child classes defined another knockback function with the previous signature, everything would work as expected).

If you're talking about this violating the substitution principle, that can be fixed with a few more argument checks.

Right now, those two are completely unused so I thought it would be a good idea to remove them. After all, nobody likes to see a func(null, 0, $x, $y).

shoghicp:
Plugins can override that. PHP does have strict checking for signatures with class checks, which happens in this case, please do read the manual.

I'm reverting it.

sekjun9878:
Okay

shoghicp:
The idea of knockBack is that it can only be applied by entities - not using it right now on PM code doesn't mean that is not going to be used in the future, and plugins might like to know the Entity pushing them.

For example, look at the MC Java code for knockback: https://github.com/Bukkit/mc-dev/blob/master/net/minecraft/server/EntityLiving.java#L713-L729

Entity is not used there but it's used by other entities. As we have not implemented AI or other stuff for other entities, of course you won't see it implemented elsewhere right now.

PR #3429, opened 2015-08-30
Fixed color in some language

MagicDroidX:


shoghicp:
To properly fix this, we need a format system and a "compiler" to colors that does not depend on putting format codes everywhere.

That is already planned, and will fix all of these issues and allow other colors that are not white

PR #3432, opened 2015-08-30
Redstone has to break 5 times to be finally mined.Attempt 2

willowmaster66:
I have changed two values to match the values in RedstoneOre.php and removed the extend to RedstoneOre and other deletes. Just the two values. :)

### DIFF
--- a/src/pocketmine/block/GlowingRedstoneOre.php
+++ b/src/pocketmine/block/GlowingRedstoneOre.php
@@ -34,7 +34,7 @@ class GlowingRedstoneOre extends Solid{
 	}
 
 	public function getHardness(){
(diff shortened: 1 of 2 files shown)


SkygridRex:
Are you using the same values of RedstoneOre.php?

willowmaster66:
I thought I did. Now you make me doubt. Do you think not?

willowmaster66:
I checked it again. These are the same values as RedstoneOre.php

willowmaster66:
I also fixed the fact that the RedstoneOre starts out as GlowingRedstoneOre. It will now become GlowingRedstoneOre after hitting it for the first time.

PR #3434, opened 2015-08-31
Delete somethings I don't need.

yeungkachun:


legoboy0215:
WTH. Why do you create a pull? You don't need it, but WE need it! Please look at what you deleted.

PR #3445, opened 2015-09-04
EntityRegainHealthEvent is not called

Creeperface01:


willowmaster66:
Looks like this pull request is not purely the EntityRegainHealthEvent fix.

Creeperface01:
ops xD how can i fix it im not too good on github

PEMapModder:
@Creeperface01 I think the event is called in heal(). https://github.com/PocketMine/PocketMine-MP/blob/f7ec1de0fd643b9543461eaf3a4e2d3c49b5dcd7/src/pocketmine/entity/Entity.php#L598

PR #3451, opened 2015-09-05
Fixed function called on Inventory not being defined

Striker209:


PEMapModder:
This is not the proper way to fix the issue.

PR #3456, opened 2015-09-07
Mcpe 0.12

Creeperface01:


shoghicp:
Non-implemented stubs or blocks are not allowed. What is the point of them?

PR #3478, opened 2015-09-13
change getOnlinePlayers to >= getMaxPlayers to align the two

LouisBHirst:
getOnlinePlayers doesn't reflect the player currently logging in so changed comparison to >=


shoghicp:
I think you want to do this on mcpe-0.12, where getOnlinePlayers() only returns the players already connected

PR #3493, opened 2015-09-15
Changing the return value of getName

0929hitoshi:


TahaTheHacker:
Instead of "EnchantTable" you can make it "Enchantment Table"

shoghicp:
Can you make everything a single commit?

0929hitoshi:
I've summarized #3495

Issue #3433, opened 2015-08-31
Compound::$item


Artide:
No new plugin or changes, just updated pocketmine and started getting so much crash dump. Heres the crash: 31.08 01:18:19 [Server] Server thread/NOTICE UndefinedPropertyException: "Undefined property: pocketmine\nbt\tag\Compound::$Item" (E_NOTICE) in "/src/pocketmine/entity/Item" at line 72

flaxues:
maybe post a CrashDump

and it think this is a duplicate of #3461

PR #3502, opened 2015-09-17
#3433 add check for Compound::$item

ecoron:
add missing check if namedTag->Item exists
should solve the issue #3433

### DIFF
--- a/src/pocketmine/entity/Item.php
+++ b/src/pocketmine/entity/Item.php
@@ -2,11 +2,11 @@
 
 /*
  *
- *  ____            _        _   __  __ _                  __  __ ____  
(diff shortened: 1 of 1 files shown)


shoghicp:
If that doesn't exist the item is invalid and should be removed,

also your commit changed formatting.

PEMapModder:
Trailing spaces should just be removed :cry: many IDEs do that

rock2rap:
How to fix it?

ecoron:
it's more to prevent the crashes if $this->namedTag->Item wasn't proper initialized.
i think the real issue is the missing initialization, but it's not clear where to do.
and yes my IDE removes trailing spaces automaticly. normaly it replace also tabs with 4 spaces ;)

shoghicp:
This is already fixed in php7 branch

PR #3505, opened 2015-09-18
Stonecutter recipes

willowmaster66:
Updated blocks/items with constants and added new block/items needed for some of the recipes.

Then removed the shapeless stone cutter recipes and added the shaped recipes. Tried to create a tabular array for the recipes with only one type of ingredient.


shoghicp:
Good work, I'll review this tomorrow.

TahaTheHacker:
Awesome :+1: :+1:

D3thw0lf:
NICE! :+1:

Galang23:
Love this change!

SkygridRex:
Lol! Why shogui don't merge this?

willowmaster66:
@SkygridRex: I think he has a lot on his mind with php7. Just be patient and let him finish it. This code won't run away. :-)

willowmaster66:
I have been thinking about moving the recipes to the class where the result of the recipe is of the type of the class. You can call a function like Stone::getRecipes() which will return all the recipes for Stone. That might be cleaner for the CraftingManger class.

What do you think?

shoghicp:
Let's not make stones (or other items) know about recipes :)

It's ok to have a way to do that on the recipe manager, but you can have blocks without recipes (it's a separated system)

PR #3506, opened 2015-09-18
Spectator can not block block placing

Creeperface01:


shoghicp:
The player itself will see the other entity (spectator) and it'll be blocked client-side

PR #3507, opened 2015-09-19
Added Player to CraftItemEvent

alejandroliu:
This add a getPlayer method to the CraftItemEvent.  This is needed to enable SimpleAuth to cancel the crafting event if the player is not yet authenticated.


shoghicp:
If crafting is done with transactions, you can block the transaction itself. Still, we should generalize this later with an interface, because maybe it won't be a player in the future (automation mods?)

Pub4Game:
Minecraft winodws 10 the action is not cancelled

PR #3512, opened 2015-09-22
Fixed crash when using Plugin::getLogger()

LDX-MCPE:
...and probably many more cases.

### DIFF
--- a/src/pocketmine/plugin/PluginLogger.php
+++ b/src/pocketmine/plugin/PluginLogger.php
@@ -93,4 +93,6 @@ class PluginLogger implements \AttachableLogger{
 			$attachment->log($level, $message);
 		}
 	}
+
+        public function __destruct() { }
 }
--- a/src/pocketmine/plugin/PluginLogger.php
+++ b/src/pocketmine/plugin/PluginLogger.php
@@ -94,5 +94,45 @@ class PluginLogger implements \AttachableLogger{
 		}
 	}
 
-        public function __destruct() { }
+        public function logException(\Throwable $e, $trace = null){
+		if($trace === null){
+			$trace = $e->getTrace();
+		}
+		$errstr = $e->getMessage();
+		$errfile = $e->getFile();
+		$errno = $e->getCode();
+		$errline = $e->getLine();
+		$errorConversion = [
+			0 => "EXCEPTION",
+			E_ERROR => "E_ERROR",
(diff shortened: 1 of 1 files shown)


PEMapModder:
Why does this cause an issue? `__destruct` is only called when there are completely no (strong) references to the object. And putting an empty method there doesn't seem to help. Can you explain more?

MagicDroidX:
Well...I dont know much but Plugin is an interface and it doesn't have a static method getLogger. It is a syntax error, i think :D

PEMapModder:
Nobody is using `Plugin::getLogger()` as a static method... We are just referring to that method, not using that code directly...

willowmaster66:
It does override the destructor in the base class.

LDX-MCPE:
```[23:53:03] [Server thread/INFO]: Loading HIDDEN v1.0b

Fatal error: Class pocketmine\plugin\PluginLogger contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Logger::logException) in phar:///root/HIDDEN/PocketMine-MP.phar/src/pocketmine/plugin/PluginLogger.php on line 27
[23:53:03] [Server thread/EMERGENCY]: An unrecoverable error has occurred and the server has crashed. Creating a crash dump
[23:53:03] [Server thread/EMERGENCY]: Please upload the "/root/HIDDEN/CrashDump_Mon_Sep_21-23.53.03-UTC_2015.log" file to the Crash Archive and submit the link to the Bug Reporting page. Give as much info as you can.
[23:53:03] [RakLibServer thread/EMERGENCY]: [RakLib Thread #139751860467456] RakLib crashed!```

LDX-MCPE:
When I used to get that error way back in PocketMine 1.3, it was because my plugin didn't have the __destruct function.

LDX-MCPE:
Nevermind that, it appears that the Interface Logger has a logException method... I'll push an update to my fork.

PEMapModder:
@LDX-MCPE it happened in 1.3 because the Plugin interface did indeed have a __destruct method.

LDX-MCPE:
@PEMapModder I realized that, that's why I pushed that commit to my pull request. I had to search what an abstract class was and learn more about interfaces.

shoghicp:
This is in my local branch already, more stuff tomorrow!

LDX-MCPE:
Thanks @shoghicp.

PR #3521, opened 2015-09-25
Added missing logException method

alejandroliu:
Adding missing logException method to PluginLogger (php7 branch, 2.0.0 API)


alejandroliu:
Checks failed because they use PHP5 whereas this branch is supposed to use PHP7

shoghicp:
Yep, this is already on my local changes. those will get pushed today!

PR #3522, opened 2015-09-25
Make sure Level::tickRateTime is (int)

alejandroliu:
tickRateTime should be an int (according to the getTickRateTime signature).


shoghicp:
The signature is wrong in this case :(

PR #3525, opened 2015-09-26
Allow escape for period in config nested keys

PEMapModder:
Expected result:

```json
{
  "a": {
    "b.c": "b period c",
    "b": {
      "c": "b slash c"
    }
  }
}
```

```php
$config = new Config("abc.json", Config::JSON);
var_dump($config->getNested('a.b\.c')); // string(10) "b period c"
var_dump($config->getNested("a.b.c")); // string(9) "b slash c"
```

Note: this mechanism isn't fully tested.

### DIFF
--- a/src/pocketmine/utils/Config.php
+++ b/src/pocketmine/utils/Config.php
@@ -259,7 +259,18 @@ class Config{
 	 * @param $value
 	 */
 	public function setNested($key, $value){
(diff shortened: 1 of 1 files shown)


jasonwynn10:
@SOF3 I think this can be used in pmmp's version

PR #3526, opened 2015-09-26
Fixed getNested() using cache desynchroinized from set()

PEMapModder:


shoghicp:
Does this also support #3525 ?

PEMapModder:
The only exceptional case I can imagine is a root key ending with a backslash, which is very unreasonable (much more dunreasonable than having keys with periods)

PR #3533, opened 2015-09-27
Fix event call when applying effects

MagicDroidX:


PEMapModder:
The event is called in Entity->attack. You don't need to explicitly call it, or it will fire twice.

MagicDroidX:
@PEMapModder Yea. You're right.

PR #3535, opened 2015-09-27
Fix return type

uniaspiex:
Fix error:
[Server thread/CRITICAL]: TypeError: "Return value of pocketmine\level\Level::getSeed() must be of the type integer, string returned in /root/src/pocketmine/level/Level.php on line 2708" (EXCEPTION) in "/src/pocketmine/level/Level" at line 2708

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -2705,7 +2705,7 @@ class Level implements ChunkManager, Metadatable{
 	 * @return int
 	 */
 	public function getSeed() : int{
-		return $this->provider->getSeed();
+		return (int)$this->provider->getSeed();
 	}
 
 	/**


shoghicp:
Why would the provider return a string in this case? What provider are you using?

uniaspiex:
I do not know, but my decision fix the issue.

shoghicp:
Yes, it fixes the issue by ignoring the original problem, a provider returning a string when THAT should return an integer. It should have a int at the start and no need to cast it.

Try to find that and submit a PR for that issue!

PEMapModder:
If you don't know the cause of the issue, then you are only silencing the bug with the possibility of leaving the bug there causing more trouble in the future, where it will become even harder to resolve the issue as there will be no errors.

CyberPowerCrazii:
my issue has more details

PR #3544, opened 2015-09-30
Added classes for items with constants

PEMapModder:


PEMapModder:
Didn't notice I'm duplicating the IDs.

willowmaster66:
I'm not sure a minecart is an item. Maybe a tile or a block.

PEMapModder:
@willowmaster66 the Minecart is not a block. It is an entity. But when you have crafted a Minecart in your inventory, it is a Minecart.

willowmaster66:
@PEMapModder: A minecart is also an item in the inventory, but it is more. It is a moving "chest". It contains players. So it can not be purely an item as you have done in this case.

PEMapModder:
@willowmaster66 yes, but a minecart item is only used as an Item in this class. For the minecart entity, it should be implemented in another class just like arrow.

PR #3570, opened 2015-10-08
Fix for nether brick fence not breaking.

willowmaster66:
Can someone please test this fix? I don't have time to test it.


Pub4Game:
It was all right, just there are no data about hell the fence in this file https://github.com/PocketMine/PocketMine-MP/blob/0.12-items/src/pocketmine/block/Fence.php or am I mistaken?

willowmaster66:
At first I thought it was a type of fence. But after checking the minecraft wiki I noticed it is a seperate object. I forgot to change that piece of code accordingly.

PR #3568, opened 2015-10-07
Fixed recursive call

MagicDroidX:

PR #3573, opened 2015-10-09
from master

wrewolf:
https://github.com/PocketMine/PocketMine-MP/pull/3568


PEMapModder:
If this is merge, it is likely that a conflict would occur when `mcpe-0.12` is merged into `master` or the other way round. Also, you are not the author of this commit.

So instead, you should create a pull request to merge `MagicDroidx:patch-3` into `mcpe-0.12` or `master` into `mcpe-0.12`, instead of creating a new commit.

PR #3579, opened 2015-10-11
Implement method getResult

MagicDroidX:


willowmaster66:
Did the server crash because this function didn't exist or what? Can you tell us why you made this pull request? Please at a better description.

MagicDroidX:
Never. But this class must implement this method. It doesnt crash just because it is not used yet.

dktapps:

PR #3580, opened 2015-10-11
Delete NetherReactor.php

ZenaGamingsky:
Because it is not in mcpe 0.12.1 and there a better neather?


TahaTheHacker:
We know.. but Nether Portals isn't in PocketMine yet.

ZenaGamingsky:
we can add it

PEMapModder:
The class exists here just to let the item be shown in chat with the proper item name instead of `Unknown`.

PR #3581, opened 2015-10-11
Delete NetherReactor.php

ZenaGamingsky:


TahaTheHacker:
We Know, but PocketMine doesn't have support for nether portals yet.

MCPEPIG:
Nether Reactors don't work in pocketmine either, @TahaTheHacker

PR #3584, opened 2015-10-11
I add new items andspawn eggs

ZenaGamingsky:


willowmaster66:
Please don't add things that are not yet implemented. For example the rails and powered rails don't work. There is also no AI for the mobs so these should not be added either.

ZenaGamingsky:
Ok how i make the AI

PR #3585, opened 2015-10-11
update all tools add on creative and add more spawn eggs

ZenaGamingsky:
Fixed bugs


willowmaster66:
Please try to create one big pull request instead of  multiple that address the same issue.

PR #3588, opened 2015-10-12
Removed nether reactor

Creeperface01:


PEMapModder:
Please do everything in the same commit.

In case you need help to checkout a non-master branch on your local repository:

```bash
git remote add base https://github.com/PocketMine/PocketMine-MP.git
git fetch base
git checkout remotes/base/mcpe-0.12
git checkout -b mcpe-0.12-fork
```

PR #3589, opened 2015-10-12
Delete NetherReactor.php

ZenaGamingsky:
Because it is not on MCPE 0.12.1 and makes lag?


ZenaGamingsky:
ok

PEMapModder:
Please stop making duplicated pull requests.

PR #3597, opened 2015-10-14
Finally signs drop..

thebigsmileXD:
Checks if the block the sign is on is updated and drops the sign if so.

### DIFF
--- a/src/pocketmine/block/WallSign.php
+++ b/src/pocketmine/block/WallSign.php
@@ -31,6 +31,13 @@ class WallSign extends SignPost{
 	}
 
 	public function onUpdate($type){
+		echo "update $this->meta\n";
+		if($type === Level::BLOCK_UPDATE_NORMAL){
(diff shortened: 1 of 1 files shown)


PEMapModder:
Please do not add debug messages into the commits

Falkirks:
Also if you want to debug use the debug method builtin to PocketMine, not `echo`.

willowmaster66:
I believe Intyre fixed this in a commit to the 0.12 blocks and items branch. It has more code.

thebigsmileXD:
@PEMapModder Oops, i'll update.
@Falkirks i don't even know why i have used echo().. i bet because i was on this file: https://github.com/alejandroliu/bad-plugins/blob/master/Mobsters/src/aliuly/mobsters/Main.php
@willowmaster66 Maybe less code is better ;)

thebigsmileXD:
This update was dumb..

PR #3598, opened 2015-10-14
Added melting, fixed snowball drop, added layers

thebigsmileXD:
I do not know if this works perfectly.
Snow layers now drop the right amount of snow balls (like pc version, one snow ball per layer)
Snow melts if block light level is too high (above 9)
You can place down multiple layers of snow (not sure if client does support this)

### DIFF
--- a/src/pocketmine/block/SnowLayer.php
+++ b/src/pocketmine/block/SnowLayer.php
@@ -50,37 +50,48 @@ class SnowLayer extends Flowable{
 		return Tool::TYPE_SHOVEL;
 	}
 
-
 	public function place(Item $item, Block $block, Block $target, $face, $fx, $fy, $fz, Player $player = null){
 		$down = $this->getSide(0);
 		if($down->isSolid()){
-			$this->getLevel()->setBlock($block, $this, true);
-
-			return true;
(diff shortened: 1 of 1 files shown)


MCPEPIG:
:+

thebigsmileXD:
What about this?
@PEMapModder

src/pocketmine/block/SnowLayer.php
```diff
@@ -50,37 +50,48 @@ public function getToolType(){
...
-			$this->getLevel()->setBlock($block, $this, true);
-
-			return true;
+			if($down->getId() === $this->getId() && $down->getDamage() <= 7){
+				$down->setDamage($down->getDamage() + 1);
+				// $this->getLevel()->setBlock($down, $down, true);
+				
+				return true;
+			}
+			else{
```
PEMapModder: Formatting >.<
But doesn't matter. I'll fix these when I merge them (if I will).

thebigsmileXD: Ou, i see them now.. Wait, i'll create a new pr after you maybe merge the new pr for the cacti + sugarcane, so it will stay clean

PR #3601, opened 2015-10-14
Update Item.php

ZenaGamingsky:
Add const BREWING_STAND = 117; in Creative for a decoration

### DIFF
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -177,6 +177,7 @@ class Item{
 
 	const NETHER_BRICKS_STAIRS = 114;
 
+        const BREWING_STAND = 117;
 	const END_PORTAL = 120;
 	const END_STONE = 121;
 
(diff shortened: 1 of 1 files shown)


PEMapModder:
Should not be in this branch.

src/pocketmine/item/Item.php
```diff
@@ -619,7 +620,6 @@ private static function initCreativeItems(){
 		self::addCreativeItem(Item::get(Item::GLASS, 0));
 		self::addCreativeItem(Item::get(Item::GLOWSTONE_BLOCK, 0));
 		self::addCreativeItem(Item::get(Item::VINES, 0));
-		self::addCreativeItem(Item::get(Item::NETHER_REACTOR, 0));
```
UltimateMcraft: What is this for? its not in the description

It looks like you are deleting the nether reactor? Why in this? You made lots of requests about this.

PR #3610, opened 2015-10-16
Fixed item damage not take effect issue.

202831820:
Fixed item damage not take effect issue.


willowmaster66:
This has already been fixed with another pull request.

202831820:
which one?

willowmaster66:
3560

Issue #3609, opened 2015-10-16
Crash on PlayerPreLoginEvent


thebigsmileXD:
The server crashes when you do
```php
public function onPlayerLogin(PlayerPreLoginEvent $event){
		$player = $event->getPlayer();
		$player->close("", "You are banned");
		$event->setCancelled();
	}
```
Crash:
Fatal error: Call to a member function setHeldItemSlot() on null in phar://C:/Users/Administrator/Desktop/Server/MinecraftPE/PocketMine-MP/PocketMine-MP.phar/src/pocketmine/Player__32bit.php on line 1772

Call Stack:
    0.0176     408808   1. {main}() C:\Users\Administrator\Desktop\Server\MinecraftPE\PocketMine-MP\PocketMine-MP.phar:0
    0.0199     403904   2. require('phar://C:/Users/Administrator/Desktop/Server/MinecraftPE/PocketMine-MP/PocketMine-MP.phar/src/pocketmine/PocketMine.php') C:\Users\Administrator\Desktop\Server\MinecraftPE\PocketMine-MP\PocketMine-MP.phar:1
    0.1494     543960   3. pocketmine\Server->__construct() phar://C:/Users/Administrator/Desktop/Server/MinecraftPE/PocketMine-MP/PocketMine-MP.phar/src/pocketmine/PocketMine.php:464
   15.3231   19858488   4. pocketmine\Server->start() phar://C:/Users/Administrator/Desktop/Server/MinecraftPE/PocketMine-MP/PocketMine-MP.phar/src/pocketmine/Server__32bit.php:1778
   15.3401   19874568   5. pocketmine\Server->tickProcessor() phar://C:/Users/Administrator/Desktop/Server/MinecraftPE/PocketMine-MP/PocketMine-MP.phar/src/pocketmine/Server__32bit.php:2178
  279.9903   19920744   6. pocketmine\Server->tick() phar://C:/Users/Administrator/Desktop/Server/MinecraftPE/PocketMine-MP/PocketMine-MP.phar/src/pocketmine/Server__32bit.php:2300
  279.9905   19920808   7. pocketmine\network\Network->processInterfaces() phar://C:/Users/Administrator/Desktop/Server/MinecraftPE/PocketMine-MP/PocketMine-MP.phar/src/pocketmine/Server__32bit.php:2485
[...]

PR #3611, opened 2015-10-16
Fix for #3609

thebigsmileXD:
Fixes crash caused by kicking a player that has no item selected (or so..)
See #3609

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1766,7 +1766,7 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 					break;
 				}
 
-				if($this->isCreative()){
+				if($this->isCreative() || $this->inventory->getHotbarSlotIndex(0) === null){
 					$this->inventory->setHeldItemSlot(0);
 				}elseif(!$this->inventory->getHotbarSlotIndex(0) === null){
 					$this->inventory->setHeldItemSlot($this->inventory->getHotbarSlotIndex(0));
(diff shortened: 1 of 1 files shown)


src/pocketmine/Player.php
```diff
@@ -1768,7 +1768,7 @@ public function handleDataPacket(DataPacket $packet){
 
 				if($this->isCreative()){
 					$this->inventory->setHeldItemSlot(0);
-				}else{
+				}elseif(!$this->inventory->getHotbarSlotIndex(0) === null){
```
PEMapModder: This isn't correct anyway. You will convert $this->inventory->getHotbarSlotIndex(0) into a boolean if you do it like this.

thebigsmileXD: By the way, if you wonder why i do so many pull requests.. I don't want to build the phar myself, because jenkins adds inprovements

PEMapModder:
THe problem in #3609 is that `Player->inventory` is null. The code here doesn't seem to fix the issue.

willowmaster66:
It has already been fixed with another pull request. Who does check the null value.

thebigsmileXD:
@willowmaster66 Your fix crashes the server

thebigsmileXD:
God thanks, this works.

thebigsmileXD:
closes #3609

willowmaster66:
3451 That's how it should be fixed. 
My fix was about every time you connect,  the first item on the hotbar is incorrect. It holds the first item in the inventory.

thebigsmileXD:
@willowmaster66 It uses your method, but with an extra check.

PR #3627, opened 2015-10-18
maybe fixed #3500

### LINKED ISSUE #3500: Uninitialized string offset: -1 File: /src/pocketmine/level/format/mcregion/Chunk__64bit Line: 121
I'm getting about 50 crashes a day with this error on 3 servers with about 40 players each on most of the time. Players are always at y<=1 when it happens, and the chunk in questions seems to contain blocks that allow users to fall through them into the void. I'm running 3 servers on maps that had no issues in 1.5, and they all have the same problem with 1.6dev builds.

If players spawn to the XYZ in question the server crashes, but when you visit the XYZ without spawning there, it doesn't crash every time, but things get strange.

CrashDump: http://pastebin.com/Xx2QmQyC
Video: of MCPE behaviour at the problem XYZ: https://youtu.be/3JwOITjPGXE

Creeperface01:


Pub4Game:
@BlindTail yes. me too helping

Pub4Game:
@Intyre You accept the changes? This is quite an important issue

PR #3651, opened 2015-10-25
Implement all effects

Creeperface01:


PEMapModder:
Branch mcpe-0.12?

Creeperface01:
why 0.12? these potion was added in 0.12?

PEMapModder:
Well, idk, but it appears that there isn't hunger until 0.12...

PR #3657, opened 2015-10-27
Arrow crash fix

hmy2001:


hmy2001:
#3657

PEMapModder:
If it is an Arrow it wouldn't be a DroppedItem.

hmy2001:
OK, I'll try to change a little.

hmy2001:
@PEMapModder

PR #3658, opened 2015-10-27
Arrow crash fix

hmy2001:


willowmaster66:
Can you explain why this solves the problem?

hmy2001:
@willowmaster66 Maybe, Shot Arrow isn't DroppedItem and ItemEntity.

hmy2001:
@willowmaster66 The Arrow which was shot by the player can not be the Item or Entity correctly.

willowmaster66:
Have you tested this? Because it looks to me as if the arrow will never disappear, because the broadcast has been deleted. The clients don't know the arrow has been added to the inventory.

PEMapModder:
@willowmaster66 line 1299/1289?

HmHmmHm:
@willowmaster66 This pull request is clearly fix client crash problem. but i think this is not wrong by pocketmine, why mc:pe 0.12 is can't apply TakeItemPacket to Arrow entity... '~'?;

willowmaster66:
@PEMapModder Kill eventually calls despawn which sends updates to the players. So that makes them disappear.
@HmHmmHm Fixing client crash is fine by me, but I'm just trying to understand if this is the correct way to go. Seems to me that removing this code might have side effects. If not I'm fine with it.

PEMapModder:
According to the behaviour in MCPE, picking up arrows doesn't seem to have anything in common with picking up items except the entity despawning and the inventory change.
Why should MCPE apply TakeItemPacket to ArrowEntity? ArrowEntity isn't an item.

HmHmmHm:
@willowmaster66 I already patched this code in my server, this code patch is really not harm.

Intyre:
But why?
Line 1289: $pk->eid = $this->getId();
Line 1294: $pk->eid = 0;
When I tested it $this->getId() was 1

HmHmmHm:
hmm...i think that reason is eid is different 'self eid'

willowmaster66:
@HmHmmHm If I take all the money from your bank account you are not hurt, but try to buy some food with no money. I'm trying to get clear if there are some side effects to deleting this code which has been added not so very long ago. Don't have a revision number right now. It has to do with players having ID 0. Or eid 0.

willowmaster66:
I have checked it. If you leave the first 4 lines and comment out the rest, it doesn't crash. If I comment out all lines and the $this->inventory and $entity->kill line, the arrow does not get picked up. If I enable only the kill line then it gets kicked up. Intyre then joined my server and confirmed that he saw the arrow getting picked up. So removing all the lines as done by this pull request has no side effects as far as I can see.

HmHmmHm:
@willowmaster66
taking my money? is it joke...?
so that's not funny bro(´･ω･`)..

and side effects is only need to
check the client side.
(because that code is client packet code)

willowmaster66:
@HmHmmHm It is a metaphor... And if you don't tell me how you tested if there are any side effects how can I know you tested it right? Or tested at all. 
Making a pull request is not as easy as just fixing the code (I had the same experience the first time I created a pull request). You also have to describe what you fixed/implemented and why.

HmHmmHm:
@willowmaster66  (´･ω･`).. oh bro...

as i said "I already patched this code in my server, this code patch is really not harm"

That mean is just i tested server side.. and
top comment as i already said...

"This pull request is clearly fix client crash problem. but i think this is not wrong by pocketmine, why mc:pe 0.12 is can't apply TakeItemPacket to Arrow ItemEntity... '~'?; is mojang fault?"

It mean i tested client side,,,
if you don't trust for me

just check you again
and if my test having a problem,

then you can comment here about that
(´･ω･`).. is it better right?

I don't think so is not difficult.

PEMapModder:
@Intyre players always get 0 for their own entity ID. 0 is a special entity ID, meaning "self".

MagicDroidX:
@PEMapModder so in short, i wonder if this PR fixes the crash currently without other issues?

PEMapModder:
I believe that this PR just acts like a mechanism where you step on another entity and that entity dies/despawns. I don't see what's wrong to addn an item to an inventory of how despawning an entity can crash clients of cause other issues.

PR #3663, opened 2015-10-31
Cleaner (and maybe faster) INI parsing at BaseLang.php

PEMapModder:


SkygridRex:
have you tested that?

PEMapModder:
@SkygridRex from the look of the code, I don't think anything could probably go wrong with that. But for your information, yes.

willowmaster66:
There is a possibility that $t[1] will fail. In the old code it wouldn't. I think you have to check if index 1 exists.

PEMapModder:
@willowmaster66 what is [this](https://github.com/PocketMine/PocketMine-MP/pull/3663/files#diff-027ad9b90052a706797368dcecb525e2R65)?

```php
 				if(count($t) < 2){
 					continue;
 				}

willowmaster66:
@PEMapModder I have overlooked that. It can not go wrong then. :)

SkygridRex:
You've heard the saying "if something works well do not touch"?

willowmaster66:
@SkygridRex If everyone would think like that nothing would improve. Saving some cpu cycles doesn't seem like a lot, but it does in the big picture I think.

PEMapModder:
It doesn't actually affect a lot in terms of performance since it is only used in startup code (for core). But still, cleaner code is always better.

PR #3671, opened 2015-11-03
Fixed typo

HelloWorld017:
Changed pocktemine to pocketmine


TahaTheHacker:
even.. there is a code break (//)..

HelloWorld017:
But it is phpdoc and IDEs use this for check which class methods return. And, it is pretty anoying that no code suggestions showing.

PR #3672, opened 2015-11-03
Update PocketMine.php

AbelGamerC:
Add to versión 0.13.0 b1


SkygridRex:
PocketMine doesn't Have all implementations needed to work with MCPE 0.13.0 close this its not only change the version you most implement all new stuff

PEMapModder:
API_VERSION is not related to MCPE versions.

PR #3707, opened 2015-11-18
Fixed hotbar has too many slots

thebigsmileXD:
The hotbar is only 6, not 9 big..

### DIFF
--- a/src/pocketmine/entity/Human.php
+++ b/src/pocketmine/entity/Human.php
@@ -98,12 +98,12 @@ class Human extends Creature implements ProjectileSource, InventoryHolder{
 
 		if(isset($this->namedtag->Inventory) and $this->namedtag->Inventory instanceof Enum){
 			foreach($this->namedtag->Inventory as $item){
-				if($item["Slot"] >= 0 and $item["Slot"] < 9){ //Hotbar
+				if($item["Slot"] >= 0 and $item["Slot"] < 6){ //Hotbar
 					$this->inventory->setHotbarSlotIndex($item["Slot"], isset($item["TrueSlot"]) ? $item["TrueSlot"] : -1);
 				}elseif($item["Slot"] >= 100 and $item["Slot"] < 104){ //Armor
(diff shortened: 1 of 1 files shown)


PEMapModder:
The hotbar size depends on the GUI size. 9 is the maximum size possible (you can see taht in win10)

thebigsmileXD:
@PEMapModder but this is the branch for 0.11. The Win10 Version is only made for 0.12 or higher

PEMapModder:
Also, the NBT structure follows the MCPC format, which is obviously 9 slots.

thebigsmileXD:
Slots that are offset are selected or can be selected. And the inventory size

thebigsmileXD:
Also, in my oppinion it should be saved into a variable. public $hotbarSize = 9||6;

iksaku:
Why should you need to `limit` players' hotbar? They're free to use whichever UI style they want! (Depending on their client screen size...)

thebigsmileXD:
hmm.. But still. Maybe thats the problems that slots that DONT EVEN EXIST in the GUI are selected. Cant find a picture right now..

PEMapModder:
@thebigsmileXD what's wrong with having an item in your hotbar but you can't see it? As long as you aren't holding it is not a problem. You are always able to move items from other hotbar slots or not in hotbar slots into your current hotbar slot using MCPE GUI.

BTW, if you adjust your D-Pad size you will have more or less slots.

thebigsmileXD:
That is now what i mean. You know the selection box around the item you currently use? Somethimes this box gliched out and selects a slot that doesn't exist right to the the three dots where you open the inventory.
https://forums.pocketmine.net/attachments/image-jpg.9255/

PEMapModder:
This is a client issue. It is impossible for PocketMine to know what the client hotbar size is, so of course we would take the maximum possible value.

thebigsmileXD:
In 0.11 its 6.

PEMapModder:
I had 7 slots since the update that let you adjust D-Pad size...

thebigsmileXD:
I think it always happens if you change gamemode and then select any block that was in your survival inventory before.

PR #3751, opened 2015-11-29
Update BrewingStand.php

hoyinm14mc:


robske110:
You should also add the specator gm! (You shouldn't be able to open it there)

PR #3761, opened 2015-12-07
Incorrect time fix

0929hitoshi:


src/pocketmine/PocketMine.php
```diff
@@ -274,7 +274,7 @@ function parse_offset($offset){
 		}
 
 		$parsed = date_parse($offset);
-		$offset = $parsed['hour'] * 3600 + $parsed['minute'] * 60 + $parsed['second'];
+		$offset = $parsed['hour'] / 3600 + $parsed['minute'] / 60 + $parsed['second'];
```
PEMapModder: It was correct.
`$offset` is in unit of seconds. `$parsed["hours"]` is in hours, and here we multiply it by `3600 seconds per hour`, making the output `$parsed["hours"] * 3600 seconds`. If it is `/ 3600`, it becomes `$parsed["hours"] / 3600 hours squared per second`.

PR #3792, opened 2015-12-21
Fix time lag

0929hitoshi:


PEMapModder:
@0929hitoshi are you sure? From what I see, this would send the number of days instead of the time in a day.

0929hitoshi:
It was re-fixed in #3942.

PR #3795, opened 2015-12-22
changed weird ban kick message

PocketKiller:
https://github.com/ImagicalCorp/ImagicalMine/commit/23f20d99f93ac97f60fe4ef93f7bd419c7be4432
and kick() function on Player.php $isAdmin.

### DIFF
--- a/src/pocketmine/command/defaults/BanCommand.php
+++ b/src/pocketmine/command/defaults/BanCommand.php
@@ -55,7 +55,7 @@ class BanCommand extends VanillaCommand{
 		$sender->getServer()->getNameBans()->addBan($name, $reason, null, $sender->getName());
 
 		if(($player = $sender->getServer()->getPlayerExact($name)) instanceof Player){
-			$player->kick($reason !== "" ? "Banned by admin. Reason: " . $reason : "Banned by admin.");
+			$player->kick($reason !== "" ? "Banned by admin. Reason: " . $reason : "Banned by admin.", false);
 		}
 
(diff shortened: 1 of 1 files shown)


Intyre:
Can you explain why you added ``false``? Only admins can use the ``/ban`` command.

PocketKiller:
look at `Player.php` on the kick function : 
```php
<?php
public function kick($reason = "", $isAdmin = true){
		$this->server->getPluginManager()->callEvent($ev = new PlayerKickEvent($this, $reason, $this->getLeaveMessage()));
		if(!$ev->isCancelled()){
			if($isAdmin){
				$message = "Kicked by admin." . ($reason !== "" ? " Reason: " . $reason : "");
			}else{
				if($reason === ""){
					$message = "disconnectionScreen.noReason";
				}else{
					$message = $reason;
				}
			}
			$this->close($ev->getQuitMessage(), $message);
			return true;
		}
		return false;
	}
?>
```
as your code in the ban command, if the player is banned without reason he'll be kicked with the message "Kicked by admin. reason: banned by admin.
and if there is a reason.. "Kicked by admin. reason: banned by admin. reason: test
because as default in kick() function isAdmin is true, and if it's true, it adds that "Kicked by admin. reason.." but with false and there is a reason, he'll be kicked as the reason without any additions.

iksaku:
Well, the current code changed doesn't makes sense, instead I would suggest to change the "Reason parsing" inside the `kick()` function, so it is more dynamic and can be adapted to the `/ban` command

Intyre:
Something like 78525e1f74e01d8c0cc?

iksaku:
@Intyre That's it ;) Thanks :3

PR #3797, opened 2015-12-23
Fix for crafting issue

onebone:
This bug is annoying. I hope this request to be merged until others make a better solution.

I don't know if this request is vaild, but it worked on my server correctly.

### DIFF
--- a/src/pocketmine/item/Item.php
+++ b/src/pocketmine/item/Item.php
@@ -1531,7 +1531,7 @@ class Item{
 	}
 
 	public final function deepEquals(Item $item, $checkDamage = true, $checkCompound = true){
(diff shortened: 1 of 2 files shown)


src/pocketmine/item/Item.php
```diff
@@ -1531,7 +1531,7 @@ public function onActivate(Level $level, Player $player, Block $block, Block $ta
 	}
 
 	public final function deepEquals(Item $item, $checkDamage = true, $checkCompound = true){
-		if($item->equals($item, $checkDamage, $checkCompound)){
+		if($this->equals($item, $checkDamage, $checkCompound)){
```
PEMapModder: :+1:

PEMapModder:
Resolved in d5f3c19

PR #3800, opened 2015-12-24
Change Item constructor visibility to protected

PEMapModder:
External callers should use `Item::get()` instead of `new Item`


PEMapModder:
This would break backwards compatibility for plugins that do this too. Although it is not recommended, it indeed is not specified in docs that it is internal.

PR #3802, opened 2015-12-24
Workaround for pthreads Collectable update + merge master

PEMapModder:
I'm not sure if this is done correctly. I hadn't tested it with the knowledge that AsyncTask is being scheduled on the server, but at least the server starts up correctly with PHP 7.0.1 + pthreads 3.1.5


onebone:
:+1:

PEMapModder:
Also merged `PocketMine:master` into this PR.

I would appreciate if anyone would help test this branch with a plugin that uses AsyncTask, to confirm that it works.

PEMapModder:
http://pastebin.com/gDHWj0Gb
This happened when I am testing with this branch. Not sure if it is caused by the changes in this PR yet.

Issue #3752, opened 2015-12-01
No stack limit on cactus


XxDXM-hippiexX:
There is no stack limit on cactus and if you stack it a few high and break the bottom it craches the server due to..

Maximum function nesting level of '100' reached, aborting!

thebigsmileXD:
Lol.. due to BlockUpdate when broken

XxDXM-hippiexX:
yuo

thebigsmileXD:
Same for sugar cane?

thebigsmileXD:
https://github.com/ImagicalCorp/ImagicalMine/commit/da15fe2ef0c90e88083fd2cfd378e1b23d3f8b78
This fixes it on Sugarcane and Cacti

thebigsmileXD:
DON'T USE MY FIX! It makes it worse .-."

PEMapModder:
I'm thinking about Level::scheduleUpdate($pos, 0)

thebigsmileXD:
Just wrote the code. :p because i saw your post on the redstone commit

thebigsmileXD:
https://github.com/ImagicalCorp/ImagicalMine/commit/3eaa032c72c7a35ad6be2ea399a06c3bc4115d03
Thanks @PEMapModder , it works just perfect.

PocketKiller:
xdebug?

thebigsmileXD:
You. This is PM. PocketMine won't disable xdebug because it shows backtrace for API + Debug for Plugins

dktapps:
This has long since been fixed on PMMP's PocketMine-MP.

PR #3806, opened 2015-12-25
Fixed sugarcane and cacti crash when too high #3752

thebigsmileXD:
Wow.. finally got this fixed.
When you placed more than 100 cacti above each other (or sugarcane) they
let the server crash due to overflow call.
Thanks @PEMapModder


PEMapModder:
Confirmed that this works?

thebigsmileXD:
Oh yea. Placed sand on bedrock layer and stacked it up to build limit. Then broke the sand and the items fell.

PEMapModder:
The head tree is contaminated with irrelevant commits. Please open a pull request without irrelevant commits.

thebigsmileXD:
I will

thebigsmileXD:
Oh, i see. This is just messed up because another pr is in the same branch.

SOF3:
See also: https://en.wikipedia.org/wiki/Tail_call

PR #3824, opened 2015-12-31
Fixed unable to swap hotbar item slots.

iksaku:
In recent MCPE versions (I don't exactly know which one) you can swap hotbar item slots. This commit adds that functionality.


thebigsmileXD:
Looks easy and cool, could this remove the annoying Windows-10 scrolling bug?

iksaku:
I'm not aware of that, as I don't use windows xD

iksaku:
Seems that anyways it worked (Client side)... So this may be useless...

PR #3826, opened 2016-01-01
add DestroyBlockParticle

0929hitoshi:


thebigsmileXD:
Will it fix the crash of MCPE if you walk over leaves? I think so..

thebigsmileXD:
There will be a short lag spike on the server when a block is broken. 5 players breaking block-> 0.5 seconds freeze.
I don't know why. Thats just a packet.

dktapps:

PR #3829, opened 2016-01-03
Add support for being able to spawn ZombiePigman just to get through end

gimilii:
to end flow of contributing to PocketMine project.  Next will add
support to spawn other Zombies, then will add movement.

### DIFF
--- a/src/pocketmine/entity/Monster.php
+++ b/src/pocketmine/entity/Monster.php
@@ -20,8 +20,52 @@
 */
 
 namespace pocketmine\entity;
-
+use pocketmine\event\entity\EntityDamageByEntityEvent;
+use pocketmine\Player;
+use pocketmine\network\protocol\AddEntityPacket;
+use pocketmine\item\Item as ItemItem;
 
 abstract class Monster extends Creature{
-
+	
+	// Add a createAddEnityPacket method so don't have to replicate this code in all of the subclasses
+	
+	protected function createAddEntityPacket(Player $player){
+		$pk = new AddEntityPacket();
+		$pk->eid = $this->getId();
+		$pk->x = $this->x;
+		$pk->y = $this->y;
+		$pk->z = $this->z;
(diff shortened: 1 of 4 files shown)


src/pocketmine/item/Item.php
```diff
@@ -877,6 +878,7 @@ private static function initCreativeItems(){
 		// TODO: Slime
 		// TODO: Spider
 		self::addCreativeItem(Item::get(Item::SPAWN_EGG, Zombie::NETWORK_ID));
+		self::addCreativeItem(Item::get(Item::SPAWN_EGG, ZombiePigman::NETWORK_ID));
 		//TODO: PigZombie
```
Samueljh1: you need to remove this todo

src/pocketmine/entity/Monster.php
```diff
@@ -20,8 +20,52 @@
...
+		$pk->x = $this->x;
+		$pk->y = $this->y;
+		$pk->z = $this->z;
+		$pk->speedX = $this->motionX;
+		$pk->speedY = $this->motionY;
+		$pk->speedZ = $this->motionZ;
+		$pk->yaw = $this->yaw;
+		$pk->pitch = $this->pitch;
+		$pk->metadata = $this->dataProperties;
+		return $pk;
```
PEMapModder: Implementation using `clone` might work better.

src/pocketmine/item/Item.php
gimilii: Removed.

src/pocketmine/entity/Monster.php
gimilii: I just copied this code from Zombie.php function spawnTo - I noticed it was the same code over and over again, so thought it would be better to create a method to hold the packet creation and place in a higher level class.  I am not sure how to use clone here.

dktapps:

PR #3832, opened 2016-01-04
Changed Entity to Player

AndreyNazarchuk:
I changed getEntity() to getPlayer() to avoid confusion.
PlayerDeathEvent should not use getEntity().


PEMapModder:
You should add a method instead of replacing the current one. The current one is here for IDEs to note that this method only returns a `Player` object. Although its declaration is redundant and backwards compatibility doesn't get affected by this pull request, you should keep it. (And it doesn't cause any more lag either)

AndreyNazarchuk:
kk
found out how to build pocketmine myself as jenkins isnt working :/

Samueljh1:
Why not just mark the old one as deprecated and add getPlayer() to fix backwards-compatability issues

PEMapModder:
The old one could be absent without implementation at all. It isn't deprecated.

You might need this in PHP 7 as overriding methods (can or cannot) change the return type signature.

PEMapModder:
Please remove the irrelevant commits in the branch.

PR #3881, opened 2016-01-16
Create BlockActivateEvent

Samueljh1:
Adds a new event that is sent when a block is activated. Eg: Flint and Steel -> TNT Block
- Also added pull request to Block Class which adds the event caller


xpyctum:
Bad practice.

Samueljh1:
why?

xpyctum:
It is not and executable php file

Samueljh1:
@xpyctum Whoops, fixed.

BlackShadow1:
You must create a BlockActivateEvent.php, not a BlockActivateEvent without extension

src/pocketmine/event/block/BlockActivateEvent.php
```diff
@@ -0,0 +1,59 @@
...
+	/** @var \pocketmine\block\Block */
+	protected $block;
+	
+	/** @var \pocketmine\item\Item */
+	protected $item;
+
+	/** @var \pocketmine\Player */
+	protected $player;
+
+	public function __construct(Block $block, Item $item, Player $player){
```
xpyctum: Maybe $player = null? 
Because caller can do this. (Block -> onActivate)

Samueljh1:
@xpyctum updated.

PEMapModder:
From what I remember, null is not a valid object in PHP to be passed into hinted parameters.

Samueljh1:
@PEMapModder really? it works fine.

PEMapModder:
never mind, just tested passing null works.

matcracker:
It's normal that doesn't work xD You don't call the event

thebigsmileXD:
Create even't but never call it?

matcracker:
It's the same that I said, if the event isn't call in the code, it doesn't work

Samueljh1:
@thebigsmileXD I even referenced the call PR: #3882

PEMapModder:
@Samueljh1 if one pull request is dependent on another, you must make them in the same pull request.

PEMapModder:
As this pull request alone is not meaningful, this is being closed.

PR #3887, opened 2016-01-18
Redstone

PEMapModder:
Adds redstone mechanism to PocketMine.

Apart from components related to entity collision and redstone torch, other components have been tested and considered to be quite ready.


PEMapModder:
For some reason I forgot why, this pull request is contaminated with irrelevant changes such as changes to SetTimePacket. Closing.

PR #3925, opened 2016-02-04
Update PocketMine.php

erkam2002:


UltimateMcraft:
MCPE 14.0.4?

SalmonGER:
php7-0.14 branch still uses build 1 protocol. So you also need to update the protocol.

Issue #3875, opened 2016-01-15
Invalid garbage Chunk given to Entity


Samueljh1:
Edited by @PEMapModder:

The issue that still exists has the following information:

> Here's another one, that happens when 2 players are killed by 1 single TNT:

> ```
[13:54:41] [Server thread/CRITICAL]: Could not tick level "world": Invalid garbage Chunk given to Entity
[13:54:41] [Server thread/CRITICAL]: pocketmine\utils\ChunkException: "Invalid garbage Chunk given to Entity" (EXCEPTION) in "/src/pocketmine/entity/Entity__64bit" at line 210
[13:54:41] [Server thread/DEBUG]: #0 /src/pocketmine/level/Level__64bit(1491): pocketmine\entity\Entity::createEntity(string Item, pocketmine\level\format\anvil\Chunk object, pocketmine\nbt\tag\Compound pocketmine\nbt\tag\Compound{.pocketmine\nbt\tag\Enum:pocketmine\nbt\tag\Enum{.pocketmine\nbt\tag\Double:609.5.pocketmine\nbt\tag\Double:16.5.pocketmine\nbt\tag\Double:-819.5.}.pocketmine\nbt\tag\Enum:pocketmine\nbt\tag\Enum{.pocketmine\nbt\tag\Double:-0.0024078552088078.pocketmine\nbt\tag\Double:0.2.pocketmine\nbt\tag\Double:0.027977234819371.}.pocketmine\nbt\tag\Enum:pocketmine\nbt\tag\Enum{.pocketmine\nbt\tag\Float:279.44812960776.pocketmine\nbt\tag\Float:0.}.pocketmine\nbt\tag\Short:5.pocketmine\nbt\tag\Compound:pocketmine\nbt\tag\Compound{.pocketmine\nbt\tag\Short:3.pocketmine\nbt\tag\Byte:1.pocketmine\nbt\tag\Short:0.}.pocketmine\nbt\tag\Short:10.})
[13:54:41] [Server thread/DEBUG]: #1 /src/pocketmine/level/Explosion__64bit(209): pocketmine\level\Level->dropItem(pocketmine\math\Vector3 Vector3(x=609.5,y=16.5,z=-819.5), pocketmine\item\ItemBlock Item Dirt (3:0)x1)
[13:54:41] [Server thread/DEBUG]: #2 /src/pocketmine/entity/PrimedTNT(132): pocketmine\level\Explosion->explodeB(boolean)
[...]

PR #3927, opened 2016-02-05
getDrops() missing line (?), incorrect return statement

0kino7ori:
Methods calling Block::getDrops() expect an array of tuples (not exactly Item objects) which in turn will be passed to Item::get() or similar methods to get the actual Item objects. Such a method is Explosion::explodeB().

But since Block::getDrops is returning an ackward array with just one tuple, the calling method will trigger an exception which logs critical messages to the server. Some users have reported this criticial message present for quite some time during gameplay.

In addition, the $item argument here is never used inside the method, which makes me believe there's something missing in the return statement, probably a line that got lost during some previous commit.

By adding this line (758), the method will return a proper array and actually make use of $item. It all seems to make sense for a method that is nothing more than a formatter.

### DIFF
--- a/src/pocketmine/block/Block.php
+++ b/src/pocketmine/block/Block.php
@@ -754,6 +754,8 @@ class Block extends Position implements Metadatable{
 		}else{
 			return [
 				[$this->getId(), $this->getDamage(), 1],
+				//By adding this line below, the method will return a proper array and actually make use of $item.
+				[$item->getId(), $item->getDamage(), $item->getCount()]
 			];
 		}
 	}


0kino7ori:
This PR might address issue #3875 TNT Issues.

PEMapModder:
$item is the item used to break the block. It is there for subclasses to use, such as for stone blocks to check if they are broken by pickaxes.

0kino7ori:
My bad! I did rush into conclusions.
Now if you are kind enough to follow on with this issue, I've done my homework.

I've checked all the calls to Item::get() which happen after a getDrops() and the only one using the token ... is in the Explosion::explodeB() method. There's a potential risk here, since the call to Item::get(...$drop) never checks if $drop is a populated unpackable array or at least contains the first argument (valid id). 

If this is the source of the critical messages, a simple check before the call would solve it:

```
if (!empty($drop))
    $this->level->dropItem($block->add(0.5, 0.5, 0.5), Item::get(...$drop));
```

Other methods in the code calling Item::get() will use a different notation, such as in BlockBreakEvent.php line 47:
```
		$drops = $player->isSurvival() ? $block->getDrops($item) : [];
		foreach($drops as $i){
			$this->blockDrops[] = Item::get($i[0], $i[1], $i[2]); 

```
Also, depending on the class, getDrops() is sometimes written to return an array of tuples ($id, $meta, $count), other times it's written to return an array of actual Item objects, such as in the BlockBreakEvent.php above.

On the other hand, a call to getDrops() is often followed by a call to dropItem(), which requires an actual Item object as argument. So when getDrops() returns a list ($id, $meta, $count), there's a need to instantiate an Item object using the static method Item::get() before calling dropItem().

Wouldn't it be easier and more fault-proof if all getDrops() in the code always returned an array of actual Item objects to begin with?

PEMapModder:
I have the same question, but since this is already implemented, in order to maintain backwards compatibility, this should be kept the same.
But whatever, the change in your pull request only adds bugs and doesn't fix any.

Could you show an example (critical message) of the bug you are trying to fix?

0kino7ori:
It's exactly the second issue reported in issue #3875 by @Samueljh1

PR #3930, opened 2016-02-06
Update BaseLevelProvider.php (php7-0.14)

64FF00:
Not sure why but it seems like the NBT thingy keeps writing the integer seed in a string format...

### DIFF
--- a/src/pocketmine/level/format/generic/BaseLevelProvider.php
+++ b/src/pocketmine/level/format/generic/BaseLevelProvider.php
@@ -87,7 +87,7 @@ abstract class BaseLevelProvider implements LevelProvider{
 		$this->levelData->Time = new IntTag("Time", (int) $value);
 	}
 
-	public function getSeed(){
+	public function getSeed() : int{
 		return $this->levelData["RandomSeed"];
 	}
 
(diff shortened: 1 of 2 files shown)


PocketKiller:
travis failed...

xpyctum:
@PocketKiller
[CRITICAL] You must use PHP >= 7.0
[CRITICAL] Please use the installer provided on the homepage.

64FF00:
@PocketKiller https://travis-ci.org/PocketMine/PocketMine-MP/builds/107432925 ```
[CRITICAL] You must use PHP >= 7.0
[CRITICAL] Please use the installer provided on the homepage.
```

@xpyctum :+1:

64FF00:
Imgur

PEMapModder:
Does this fix the issue of returning string? (And are you sure that it really does?) According to [php.net](http://php.net/manual/en/functions.returning-values.php#functions.returning-values.type-declaration), if strict typing is enabled, a `TypeError` would be thrown if the return value is not an int.

64FF00:
@PEMapModder Unfortunately it doesn't. I know it's just a workaround for it, but I'm sure there's something wrong with the nbt tag parsing...

PR #3945, opened 2016-02-14
Create SkullBlock.php

ZenaGamingsky:
this is SkullBlock.php so we be able to have skulls in mcpe soon


Georggi:
Can you just scroll down a bit? There is already pull request #3928 which adds it.

PR #3946, opened 2016-02-14
Fixes a bug with the slots

Pub4Game:
image

### DIFF
--- a/src/pocketmine/Player.php
+++ b/src/pocketmine/Player.php
@@ -1798,7 +1798,8 @@ class Player extends Human implements CommandSender, InventoryHolder, ChunkLoade
 				$this->setNameTag($this->username);
 				$this->iusername = strtolower($this->username);
 
-				if(count($this->server->getOnlinePlayers()) >= $this->server->getMaxPlayers() and $this->kick("disconnectionScreen.serverFull", false)){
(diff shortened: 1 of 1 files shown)


PEMapModder:
Does your image explain anything at all?

Pub4Game:
What's wrong?

Pub4Game:
I don't understand what you mean... I have everything working perfectly

PEMapModder:
@Pub4Game why did you use `close()` rather than `kick()`?

PEMapModder:
I'm unable to reproduce the issue you reported.

It works with 0/0

Pub4Game:
Okay, I will fix it

PEMapModder:
@Pub4Game fix what? Did you not see my image? This issue cannot be reproduced using the latest master build.

I have also tested with one player:



This shows that the server full mechanism is working correctly.

Pub4Game:
@PEMapModder 

I used 'close()' because if you use the 'kick()', then the following happens
image

PEMapModder:
@Pub4Game without plugins? And are you using official PocketMine?

Pub4Game:
Yes..

PEMapModder:
@Pub4Game I am unable to reproduce that issue using the latest development build.

PR #3955, opened 2016-02-18
Fix the server crash when player spawn back to Y coord. >0 & <1

redcrab2016:
If a player is creative he can mine the lowest bedrock level . Then after If the player go into the void leave and go back in the game then the server crash due to a getBlockId with Y coordinate below 0 (-1).


PEMapModder:
Please follow the [contribution guidelines](../blob/master/CONTRIBUTING.md#contributing-code):
> **No merge commits are allowed**

PR #3956, opened 2016-02-18
PR: Bug Fix - Crash if player is spawn to Y coord. >0 & <1

redcrab2016:
**Original issue:**
When player is in creative mode and mine the lowest bedrock and fly at the bedrock level (Y coordinate above 0 and lower than 1) 
then after if the player quit the server and come back, the server crash on getBlockId function with a Y coordinate at -1.

**Fix:**
 in level/level.php , getSafeSpawn is modified in order to consider that  player spawn Y location below 1  is  0 ,then use current level SpawnLocation.


robske110:
Good

PR #3957, opened 2016-02-18
Timeout connect in server.properties

abler98:


PEMapModder:
Looks more appropriate to be in `pocketmine.yml`.

PR #3973, opened 2016-02-21
PR-bugfix : getSeed expect an int and not a string (level provider issue)

redcrab2016:
**issue:**
crash dump with
Error: Return value of pocketmine\level\Level::getSeed() must be of the type integer, string returned
/src/pocketmine/level/Level
Line: 2708

**Fix:**
Update _BaseLevelProvider.php_ to be sure that getSeed() return an integer.


redcrab2016:
still travis using php5 instead of php7...

PEMapModder:
Please stop duplicating the same pull request.

PR #3997, opened 2016-02-25
Boat can be placeable

ondraczsk:
Update boat


PEMapModder:
Class `pocketmine\entity\Boat` doesn't even exist. Did you copy code?

PR #4007, opened 2016-02-29
Added more mobs, almost all!

matcracker:
Added almost all standard mobs classes and egg spawing.


LDX-MCPE:
There's really no need for this until AI is implemented, it will be easier to add each mob as the AI is added.

robske110:
AI won't (hopefully) be added. Its laggy and stupid
We just don't need it

BlackShadow1:
I think AI lag with php, php7 too, so i agree with @robske110

matcracker:
But this pull doesn't add mobs AI -.-

DiegoGTRatty:
Lag isnt really an issue when it comes to the AI...

Maybe setup a "gamemode" where the game wont spawn mobs at all?

VonHirsch:
@DiegoGTRatty https://github.com/PocketMine/PocketMine-MP/blob/master/src/pocketmine/resources/pocketmine.yml#L138

legoboy0215:
XD Plugins can add em.

PR #4008, opened 2016-02-29
Create 1337.ini

thedeibo:
All whats needed left is to add it to the InstallerLang.php and were ready to go :)


64FF00:
I really need this

PEMapModder:
translate.PocketMine.net?
If you want to add a language, discuss it on IRC instead of through pull requests - we need to add them in the website as well.

thedeibo:
I already did. ( via query ( private message ) )

PR #4011, opened 2016-03-01
Update Level.php

### LINKED ISSUE #3991: Could not tick level
[Server thread/CRITICAL]: Could not tick level "Spawn": mt_rand() expects parameter 1 to be integer, float given
[Server thread/CRITICAL]: TypeError: "mt_rand() expects parameter 1 to be integer, float given" (EXCEPTION) in "/src/pocketmine/level/Level__32bit" at line 872

Samueljh1:
Fix #3991


PEMapModder:
You should be casting `$randRange` to int in all usages. Better do it in assignment.

PR #4015, opened 2016-03-03
Added an abstract Food class

legoboy0215:
Now you can use 
`$food instanceof Food` 
;)

### DIFF
--- a/src/pocketmine/item/Apple.php
+++ b/src/pocketmine/item/Apple.php
@@ -22,7 +22,7 @@
 namespace pocketmine\item;
 
 
-class Apple extends Item{
+class Apple extends Food{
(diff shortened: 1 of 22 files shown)


KnownUnown:
There doesn't seem to be any advantages in abstracting Food away from Item. There's no fields that Food needs that Item can't have.

PEMapModder:
Well, there is a purpose. But in that case, I think calling isFood() is better for that purpose.

If you really think that a separate class is necessary, you should probably add other functions to it to make it have the purpose of an abstract class.

legoboy0215:
Oh, ok. I just though in the future, you could just add more functions like getRestoredSaturation. But I saw your commits. It seems to include that, so thanks :)

PR #4016, opened 2016-03-03
Added Materials.

PocketKiller:
It is useful for plugin API.
Question: should I add items that needs to be crafted with another item?


PocketKiller:
What did I do?!

PR #4024, opened 2016-03-06
Fixed the error "BlockHash must type of Interger"

DELOS-phardev:
###Must Type Of Interger !
(int)
(int)
(int)
(int)
(int)(int)(int)(int)

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -651,7 +651,7 @@ class Level implements ChunkManager, Metadatable{
 		$this->timings->doTickPending->startTiming();
 		while($this->updateQueue->count() > 0 and $this->updateQueue->current()["priority"] <= $currentTick){
 			$block = $this->getBlock($this->updateQueue->extract()["data"]);
-			unset($this->updateQueueIndex[Level::blockHash($block->x, $block->y, $block->z)]);
+			unset($this->updateQueueIndex[Level::blockHash((int)$block->x, (int)$block->y,(int) $block->z)]);
 			$block->onUpdate(self::BLOCK_UPDATE_SCHEDULED);
 		}
 		$this->timings->doTickPending->stopTiming();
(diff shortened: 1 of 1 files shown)


DELOS-phardev:
It is an Fatal Error !

PEMapModder:
First we have to find out why $pos->x would be a non-integer. The problem should be fixed from the source.

DELOS-phardev:
@PEMapModder Plugin Developers cannot use "getLevel()->getBlock()" method now , because of this Error

PEMapModder:
@DELOS-phardev plugin developers should not have passed a floating point vector to getBlock() in the first place, right?

robske110:
Hey! Wait this should be float

robske110:
@PEMapModder They should do that and do that!
As example if you want to get an block below the player you subtract -1 from y and then pass the FLOAT values to the getBlock

PEMapModder:
Now? This is not a real issue, but adds convenience to developers.

robske110:
Hey $pos->floor will not work with negative numbers

PEMapModder:
@robske110 negative numbers have nothing to do with int/float.

robske110:
But floor() won't work on negative values

PEMapModder:
@robske110 it does. `(int) floor(-0.5)` results in -1, which is correct.

robske110:
It's not
when you are on pos -0.5 you are on block 0 not -1!!!!!!

robske110:
```php
        function truncate_float($number, $behindNumber) {
            $amp = pow(10, $behindNumber);
            if($number > 0){
                return floor($number * $amp) / $amp;
            }else{
                return ceil($number * $amp) / $amp;
            }
        }
```

PEMapModder:
From PHP.net:

> Returns the next lowest integer value (as float) by rounding down value if necessary.

We want to round down -1.5 to -2, not -1. And that's what `floor()` does.

```shell
$ php -r 'var_dump(floor(-1.5), (int) floor(-1.5));'
float(-2)
int(-2)
```

and this is the correct behaviour. We want to round down, not to round to the integer closer to 0.

DELOS-phardev:
Hmm.. Then .. Let's Blame ShoghiCP !
@PEMapModder

ZINGDING:
@DELOS-phardev 
That's Good!

robske110:
Blame me

PR #3978, opened 2016-02-22
Implemented attributes

PEMapModder:
- [x] basic attributes system
- [x] hunger mechanism
- [ ] experience mechanism
- [x] effects
  - [x] night vision
  - [x] blindness
  - [x] hunger
    - [x] declaration
    - [x] implementation
- [x] Fix speed-related effects
- [ ] exhaustion
  - [x] attack
  - [x] damage
  - [x] break blocks
  - [ ] walk per meter: 0.01
  - [ ] sneak per meter: 0.005
  - [ ] swim per meter: 0.015
  - [ ] sprint per meter: 0.1
  - [ ] jump: 0.2
  - [x] regen per halfheart | food >= 18: 4.0
- [x] events
  - [x] exhaust event
  - [x] eat event
- [x] data saving

Resolves:
* #3922
* #3830 
* #3809
* #3412
* Possible:
  * #3894

PR #4027, opened 2016-03-07
Update Attribute.php

JackNoordhuis:
Updates the attributes file to allow sending the newest attributes, will help fix speed effect and allow developers to set custom player speeds when attributes are properly implemented.


PEMapModder:
Duplicates #3978.

PR #4060, opened 2016-03-11
Added Block::fromString() method

LDX-MCPE:
I believe this will work, based off of Item.php.


PEMapModder:
I think using `Item::fromString()->getBlock()` is still a better approach, because names like `BED` should resolve into a bed block rather than null/air. Some items have `$this->block` set without being a BlockItem.

PR #4067, opened 2016-03-13
Added settings.incompatible-plugins in pocketmine.yml

PEMapModder:
After showing a warning in output, load the plugins with incompatible API versions if `settings.incompatible-plugins` is enabled in `pocketmine.yml`


Subtleoliver:
Yass!

PEMapModder:
Shall be replaced by a better implementation.

PR #4085, opened 2016-03-22
Allows @notscript tag in script plugins

PEMapModder:
This allows files in the plugins folder to contain `.php` files that contain `/**` but are not script plugins.

### DIFF
--- a/src/pocketmine/plugin/ScriptPluginLoader.php
+++ b/src/pocketmine/plugin/ScriptPluginLoader.php
@@ -98,6 +98,10 @@ class ScriptPluginLoader implements PluginLoader{
 				$key = $matches[1];
 				$content = trim($matches[2]);
 
+				if($key === "notscript"){
+					return null;
+				}
+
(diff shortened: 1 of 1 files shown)


thedeibo:
Can you explain what you mean? 
(I don't quite understand)

src/pocketmine/plugin/ScriptPluginLoader.php
```diff
@@ -105,7 +109,7 @@ public function getPluginDescription($file){
 				break;
 			}
 		}
-		if($insideHeader){
+		if($insideHeader)){
```
PEMapModder: What the...

PEMapModder:
@thedeibo I may have a .php file in the plugins folder, but I don't want PocketMine to attempt to load it as a script plugin.

thedeibo:
Ahhh. Ok.

PEMapModder:
Rebased the commits.

PR #4092, opened 2016-03-23
Rename CONTRIBUTING.md to .github/CONTRIBUTING.md

PEMapModder:
This makes the root directory cleaner.


Falkirks:
See #4093

SOF3:
This may break backwards compatibility for tools that rely on CONTRIBUTING.md to... uhh... understand the PocketMine code style, that's the excuse.

dktapps:
This can also be used in its current form on other providers such as GitLab or Bitbucket. As such, this doesn't make sense because it is general contribution guidelines, not specific to GitHub.

PR #4096, opened 2016-03-27
To fix #4095

### LINKED ISSUE #4095: Placing blocks bug
Hey.In Gamemode creative the blocks cannot be placing.....and they disappear from hotbar...fix?This is happening only in gmc.in gms it's work perfect(Soory for my english,i'm romanian)

PocketMine:1.8
OS:Windows

Pub4Game:


src/pocketmine/level/Level.php
```diff
@@ -1756,7 +1756,11 @@ public function useItemOn(Vector3 $vector, Item &$item, int $face, float $fx = 0
 
 			Tile::createTile("Sign", $this->getChunk($block->x >> 4, $block->z >> 4), $nbt);
 		}
-		$item->setCount($item->getCount() - 1);
+		if ($player != null && $player->isCreative()) {
+			$item->setCount($item->getCount());
```
PEMapModder: What's the point of this?

Pub4Game: When you put the blocks, they are reduced by 1

Blocks in creative should not end

PEMapModder: Then what's the point of setCount() to its original value?

Pub4Game: It seems to me that the client should communicate the number of blocks

PEMapModder: https://github.com/PocketMine/PocketMine-MP/blob/master/src/pocketmine/item/Item.php#L1038

Now, do you expect simply running `$this->count = $this->count;` has a point at all?

Pub4Game: Okay :D

Pub4Game: Now, all right?

PEMapModder:
#4095 mentioned about bugs placing blocks in creative mode. This pull request contains code that is only run if player is not in creative mode. How would this fix any relevant bugs?

Pub4Game:
So I got it wrong, but this patch brings some meaning, if the creative menu was fixed before the update 0.14.0

image

PEMapModder:
I cannot reproduce the suggested issue.

Pub4Game:
@PEMapModder join to server and open creative inventory... The entire inventory will be filled with blocks, and actually should be empty

PR #4112, opened 2016-04-06
Support for 0.14.1 and 0.14.0

flaxues:


PEMapModder:
Explain the reason behind retaining `CURRENT_PROTOCOL`. Also, unexpected behaviour if something is not included in `ACCEPTED_PROTOCOLS` but supported in `CURRENT_PROTOCOL`. There should be an assertion. And possibly incorrect behaviour when a client joins with an outdated version or too-updated version in the future.
Furthermore, if there is a protocol version bump, there are usually protocol changes. Make sure you updated them.
And your changes on the MCPE version is invalid.
It appears that your changes come from another source without acknowledgement.

src/pocketmine/network/protocol/Info.php
```diff
@@ -31,6 +31,7 @@
 	 * Actual Minecraft: PE protocol version
 	 */
 	const CURRENT_PROTOCOL = 45;
+	const ACCEPTED_PROTOCOLS = [45, 46];
```
Herofied1180: Nice!

PrimusLV:
What if PocketMine sends packet from 0.14.1 to client who has 0.14.0 ?

PR #4115, opened 2016-04-06
REAL UPDATE TO 0.14.1

TheDeibo:
These changes has been agreed with a official PocketMine staff, and it does work.


PEMapModder:
There are other protocol changes apart from the protocol version bump.

TheDeibo:
Ok. (So what your saying is, I just supplied a way for the server to allow game-play, but a few protocols different?)

Lonebuddha:
I've applied the versions bump at my server, children are playing, and I do not see any issues so far.

@PEMapModder, can you share your source of information about the protocol differences (I am a newbie in MineCraft dev world).

TheDeibo:
You dont need it. :) only people who wish to do a pull request or a pocketmine dev would want it

PR #4119, opened 2016-04-09
This would 'fix' some useless issues

robske110:
That say: jenkins is down, where can i grab a new build


TheDeibo:
WHyyyyyyyyyy

PEMapModder:
Although apparently hopeless, the Jenkins server is still considered as "momentarily down" and "will be up very soon", which do not require labelling as down

robske110:
Ok, but it would realy nice if you add the link to binary!

PR #4130, opened 2016-04-16
Update Block.php

PrimusLV:
Added DestroyBlockParticle on block break


Pub4Game:
https://github.com/PocketMine/PocketMine-MP/blob/master/src/pocketmine/Player.php#L2280

Check function useBreakOn()
https://github.com/PocketMine/PocketMine-MP/blob/master/src/pocketmine/level/Level.php#L1473

> public function useBreakOn(Vector3 $vector, Item &$item = null, Player $player = null, bool **$createParticles** = false) : bool{

PR #4133, opened 2016-04-17
Fix undefined method RCONInstance::kill()

SuperMarcus:
Since pthreads no longer support Thread::kill() and Thread::detach(), remove $this->workers[$n]->kill() in RCON.php to prevent exceptions in order to save levels properly. Also, Server::microSleep(50000) will be called after the loop.


robske110:
#4157 is better

Issue #4127, opened 2016-04-16
MCPE 0.14.1 not supported


skywalker-z:
logged out due to Outdated server!

jimtbob:
This issue was brought up before and closed, I installed PocketMine-MP 1.6dev-23 phar per the previous issue - still can't get onto the server (MCPE v0.14.1 on iOS or Android says outdated server)

jellyfishcoder:
Here is a phar http://pmt.mcpe.me/pmb/dl/pr:4115.phar Its from a closed pull request but it works. They denied the pull request because they said it didnt make all the changes for 0.14.1 BUT ATLEAST IT ALLOWED IT TO BE SUPPORTED.

jellyfishcoder:
Just realised that is old link. You have to compile it yourself. https://github.com/PocketMine/PocketMine-MP/pull/4115 is the pull request. Ill upload a phar when I get a chance.

skywalker-z:
@jellyfishcoder  As you pointed out the link no longer works. Can you upload a phar please? Kids are waiting ... Or can you share instructions about how to compile the pull request? Thanks.

jellyfishcoder:
@skywalker-z Here it is: https://drive.google.com/file/d/0B0wvRQEqnhujN1k4eTBBbFhHWmc/view?usp=sharing Be sure to rename it to PocketMine-MP.phar before usingit.

jellyfishcoder:
Also the water bucket is missing in creative in it.

jellyfishcoder:
When I figure out how to compile PocketMine from the source into a phar I will put together a bunch of pullrequests and and make a phar with all the features I can fix in it. Ive found fix for the creative water bucket and for some other issues, and made them into one resp, but dont know how to compile a phar,.

skywalker-z:
@jellyfishcoder Thanks for the help, but the phar file doesn't work for me. Any idea?

.\start.cmd
Warning: require_once(phar://C:\PocketMine-MP\PocketMine-MP.phar/src/pocketmine/PocketMine
.php): failed to open stream: phar "C:\PocketMine-MP\PocketMine-MP.phar" SHA1 signature co
uld not be verified: broken signature in C:\PocketMine-MP\PocketMine-MP.phar on line 1
Fatal error: require_once(): Failed opening required 'phar://C:\PocketMine-MP\PocketMine-M
[...]

PR #4136, opened 2016-04-21
0.14.1 Update For client and server

Herofied1180:
In this pull request I will be updating the PocketMine Server **MINECRAFT_VERSION** (Which will be the MC version the server accepts) & **MINECRAFT_VERSION_NETWORK** (Which will be the MC version the server runs on). It will make MC version 0.14.1 compatible with this server and, MC 0.14.1.
It will also solve the problem in the issue forum thread #4127. That is about all I can think of that I can say, so have a great day enjoying PocketMine in MC 0.14.1! ;)

### DIFF
--- a/src/pocketmine/PocketMine.php
+++ b/src/pocketmine/PocketMine.php
@@ -72,7 +72,7 @@ namespace pocketmine {
 	use pocketmine\utils\Utils;
 	use pocketmine\wizard\Installer;
 
-	const VERSION = "1.6dev[";
+	const VERSION = "1.6dev";
 	const API_VERSION = "2.0.0";
 	const CODENAME = "[REDACTED]";
 	const MINECRAFT_VERSION = "v0.14.1.0 alpha";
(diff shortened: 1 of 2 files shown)


extremeheat:
You can't simply just change some strings and expect for things to work. This PR will not allow 0.14.1 clients to join -- you need to change the protocol version for that.

Furthermore, before you commit a PR you aught to have a bit of familiarity with the project that you're contributing to... let alone follow the contributing guidelines.

There is no good reason to add messages like "join with x client version" to the server log, it is unnecessary information.

Herofied1180:
I am sorry, I will get right to it later but right now I have other things to worry about so I will lock this until tommorrow. Good Night everyone!

Herofied1180:
Good Morning peeps! It's very early for me but i might be able to code a little.

Herofied1180:
I am editing the Network Protocol immidently

Herofied1180:
Ok does anyone know what the network protocol (e.g. MC protocol version) should be?

Herofied1180:
Ok doing Major string bug fix

Herofied1180:
Finally success! I can't wait to try this out! ;) Enjoy the PocketMine update to 0.14.1!

extremeheat:
This still does not resolve any of the issues I stated...

You still have the unneeded console messages, changes to irrelevant things, and a whole bunch of poorly named commits which should be squashed.

Have you looked into the protocol changes either before you updated the protocol version? Your PR looks just like #4115 which was closed for not accounting for protocol changes.

PR #4140, opened 2016-04-22
REAL 0.14.1 UPDATE with the latest protocol version

Herofied1180:
**REAL 0.14.1 UPDATE with the latest protocol version**
-**Protocol bug fixes**
-**Removal of unnecessary strings of text**
-***Actually*** **updates to 0.14.1**
-**Gets rid of fake updates**


Herofied1180:
Ok I have looked at the supported versions and I think that the latest protocol version for 0.14.1 is either 46 or 54. Please help me. I really want a server for 0.14.1.

Dev-Kevo:
46

PR #4150, opened 2016-04-25
Added getter for speed

luca28pet:


Intyre:
What is the reason behind this?
```php
$speed = $player->speed;      // works fine
$speed = $player->getSpeed(); // works too but why?
```

robske110:
Because we then can actually. Erm. Er. Hmm. 
Ah! Make the speed variable private

luca28pet:
Didn't see $player->speed was public

luca28pet:
Holy crap you're so funny rob

robske110:
That ironic smell
And now @Intyre will lock it and edit this comment, because i wrote this

PR #4157, opened 2016-04-29
fix crash at server stop when RCON is enable

3do2:
[17:10:36] [Server thread/INFO]: [CONSOLE: Stopping the server]
[17:10:36] [Server thread/EMERGENCY]: Crashed while crashing, killing process
[17:10:36] [Server thread/EMERGENCY]: Error: Call to undefined method pocketmine\network\rcon\RCONInstance::kill()

### DIFF
--- a/src/pocketmine/network/rcon/RCON.php
+++ b/src/pocketmine/network/rcon/RCON.php
@@ -71,7 +71,7 @@ class RCON{
 		for($n = 0; $n < $this->threads; ++$n){
 			$this->workers[$n]->close();
 			Server::microSleep(50000);
-			$this->workers[$n]->kill();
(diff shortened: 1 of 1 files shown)


extremeheat:
Squash the commits, you have two with the same changes.

robske110:
But this is better

3do2:
This fix tries to fix the same issue (worlds are probably not properly saved at server stop). I've called the quit() method of the thread class. Kill method doesnt exist anymore. I don't move the microSleep because I don't really know why it's there ? Or if it's better outside the loop.

dktapps:
The changes in this PR have been merged to the PMMP repository.

PR #4194, opened 2016-06-05
Create PlayerHungerEvent.php

sero583:
I'm not sure, but isn't this missing anyways?

### DIFF
--- a/src/pocketmine/event/player/PlayerHungerEvent.php
+++ b/src/pocketmine/event/player/PlayerHungerEvent.php
@@ -1,0 +1,43 @@
+<?php
+/*
+ *
+ *  ____            _        _   __  __ _                  __  __ ____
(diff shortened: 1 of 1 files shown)


extremeheat:
What's the point of creating this useless event?

sero583:
@extremeheat That event is called, if the hunger from the player changes, i don't think, it's useless at all.

extremeheat:
Where is it called?

sero583:
@extremeheat I don't know what you trying to tell me, how do you mean 'Where is it called'?

extremeheat:
You said that the **event is called when the hunger from the player changes**. I'm asking where in the code this event is called.

sero583:
@extremeheat I don't know, anyways someone should correct this event a bit xD

PR #4199, opened 2016-06-08
Create sign tile like it should be.

thebigsmileXD:
Every other tile is created like that. Signs go blank less frequently.
Also fixes a bug when the breakblockevent of the block below is
cancelled and a sign post is on top both are getting removed.


Muqsit:
Will this work with \pocketmine\block\WallSign

thebigsmileXD:
WallSigns extend SignPosts, so sure it does. @Muqsit #knowthecode

dktapps:
The changes in this PR have been merged to the PMMP repo (more than a year ago in fact).

PR #4200, opened 2016-06-08
Cacti + Sugarcane

thebigsmileXD:
Fixing crashes when the plants are too high and broken (without xdebug
no crash but server freeze) and some logic fixes. Cacti + Sugarcane do
grow naturally onTick, but using Bonemeal on sugarcane will not be sendt
by the client (MCPE).
https://bugs.mojang.com/browse/MCPE-15102

### DIFF
--- a/src/pocketmine/block/Cactus.php
+++ b/src/pocketmine/block/Cactus.php
@@ -28,7 +28,6 @@ use pocketmine\event\entity\EntityDamageEvent;
 use pocketmine\item\Item;
 use pocketmine\level\Level;
 use pocketmine\math\AxisAlignedBB;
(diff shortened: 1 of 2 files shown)


thebigsmileXD:
#3752 #3806

jasonwynn10:
Does this apply to pmmp's PocketMine?

thebigsmileXD:
Closed because of ban + outdated unformatted code.

thebigsmileXD:
@jasonwynn10 in theory the problems were "reduced", these loops still appear in pmmp. Ask dktapps about it. The growth script worked though, but things will change to allow different tick times for ticking blocks.

PR #4226, opened 2016-06-26
Fix a horrible entity-related memory leak

dktapps:
This commit, simple though it looks, fixes ~~the~~ **a** horrible memory leak that PocketMine and all its forks of have been suffering from since forever.

EDIT: Further explanation is probably warranted.

The problem here, as briefly as I can summarize it, is that entities are leaked on chunk unload.

**Before patch:**
- Chunk is unloaded and saved. All entities on the chunk are saved. Level::removeEntity is called on all entities in that chunk to remove them.
- Level::removeEntity kills non-player entities. Result is items being spawned in an unloaded chunk.
- Movement updates to these entitities (falling checks etc.) trigger requests to the blocks around the entities. Because the entities are in unloaded chunks, these requests cause the chunks to be reloaded. When the chunk is reloaded, the saved entities are spawned again.
- Process repeats, leaking more and more entities until memory limit is exhausted.

**After patch:**
- Chunk is unloaded and saved. All entities on the chunk are saved. Level::removeEntity is called on all entities in that chunk to remove them.
- Because Level::removeEntity now uses close() instead of kill(), no items are dropped, which prevents the leak from beginning.

### DIFF
--- a/src/pocketmine/level/Level.php
+++ b/src/pocketmine/level/Level.php
@@ -2409,7 +2409,7 @@ class Level implements ChunkManager, Metadatable{
 			unset($this->players[$entity->getId()]);
 			$this->checkSleep();
 		}else{
-			$entity->kill();
+			$entity->close();
 		}
 
 		unset($this->entities[$entity->getId()]);


SOF3:
Just to make sure that this works, any proof that this fixes **the** memory leak problem, please? I mean, for example, is this the **only** source of memory leak?

{{citation needed}}

dktapps:
@SOF3 no. This isn't the only leak. However, it was the main one, at least in Genisys. 

PM may not suffer as badly as Genisys did as far as I can tell PM mobs don't drop items or exp. But this _did_ have drastic effects. Most Genisys users enjoyed an increase in server uptime from hours to weeks.

dktapps:
We've only discovered since this was fixed that there are other leaks.

dktapps:
I've reworded the commit and updated the title since THE leak is not accurate as stated above.

thebigsmileXD:
Even a way better explaination than on the forks before, nice!

robske110:
?

dktapps:
@robske110 I was told I was wasting my time. Check Telegram

PR #4244, opened 2016-07-09
Fix .mailmap double names

Wqrld:
I dont know if this was needed but its fixed


dktapps:

PR #4254, opened 2016-07-28
Add support for 0.15.4

TheAppleGamer:


extremeheat:
Based on your PR, there's been *absolutely* no protocol changes, am I right?

TheAppleGamer:
Only one protocol change

extremeheat:
Which is? Shouldn't your PR address the change then?

dktapps:
Wooo, you guessed the magic number. Great. Now, what _actually_ changed? Version numbers don't get changed for no reason.

jojoe77777:
> Version numbers don't get changed for no reason.

I think it may just be because of the new skin pack.

dktapps:
@jojoe77777 I was referring to the protocol version, not the game version. I'll update my earlier comment

jojoe77777:
@dktapps Yes, I was also referring to that. I was wondering what would happen if someone with a new skin could join a sever with older clients... The older clients wouldn't be able to render the skin properly, so it would need a protocol bump so that doesn't happen. There may also be another reason for the bump.

extremeheat:
@jojoe77777 custom skins can already be used, so I don't really see what you're referring to here. Unless massive things changed with the skin rendering model, or how skins are transmitted, there would be no protocol-related changes.

robske110:
Well the client wouldn't be able to render these skins in the old version

dktapps:
@extremeheat is right. Thank you, you beat me to it.

PR #4260, opened 2016-08-10
PSR-2 violation

SOF3:
I don't need to give examples where this principle is violated, do I?
https://github.com/PocketMine/PocketMine-MP/search?utf8=%E2%9C%93&q=public+abstract+function&type=Code


robske110:
maybe it was planned to violate 'em?

SOF3:
@robske110 the contribution guidelines meant to follow the PSR-2 apart from the specified exceptions, but it comes out that this rule is not followed either, so it should be written out.

Furthermore, sometimes it is followed and sometimes it is not followed, so I used SHOULD rather than MUST.

dktapps:

PR #4275, opened 2016-08-27
add Blast resistance to all blocks

0929hitoshi:


robske110:
You may want to add a stub function which is returning either -1 or 15

dktapps:

